PR #11, opened 2022-10-30
[+] Plugin loader feature

hykilpikonna:
1. Create plugin loader boundary interfaces
2. Create PluginLoader implementation that hot load/unload/reload plugins while the server is running
3. Create ArgsParser and Controller to parse & execute commands
4. Create ReflectionUtils: Use reflection to read and write to private fields with generic typing
5. Create Sugar: Array slicing functions

### DIFF
--- a/src/main/java/org/hydev/mcpm/Controller.java
+++ b/src/main/java/org/hydev/mcpm/Controller.java
@@ -1,0 +1,80 @@
+package org.hydev.mcpm;
+
+import net.sourceforge.argparse4j.inf.ArgumentParserException;
+import org.hydev.mcpm.client.injector.*;
+
+import java.util.function.Consumer;
+
+import static java.lang.String.format;
+import static org.hydev.mcpm.utils.Sugar.sub;
+import static org.hydev.mcpm.utils.Sugar.subFrom;
+
+
+/**
+ * Controller of the program
+ *
+ * @author Azalea (https://github.com/hykilpikonna)
+ * @since 2022-10-30
+ */
+public class Controller
+{
+    private final LoadBoundary loader;
+    private final UnloadBoundary unloader;
+    private final ReloadBoundary reloader;
+
+    public Controller()
+    {
+        var pl = new PluginLoader();
+        loader = pl;
+        unloader = pl;
+        reloader = pl;
+    }
+
+    /**
+     * Run a command
+     *
+     * @param args Command-line arguments (not including the command name/path)
+     * @param log Callback logger
+     */
+    public void runCommand(String[] args, Consumer<String> log)
+    {
+        if (args.length == 0) return;
+
+        var cmd = args[0].toLowerCase();
+        switch (cmd)
+        {
+            case "load", "unload", "reload" ->
+            {
+                try
+                {
+                    var a = ArgsParser.LOADER.parseArgs(subFrom(args, 1));
+                    var name = a.getString("name");
+
+                    try
+                    {
+                        switch (cmd)
+                        {
+                            case "load" -> loader.loadPlugin(name);
+                            case "unload" -> unloader.unloadPlugin(name);
+                            case "reload" -> reloader.reloadPlugin(name);
+                        }
+                        log.accept(format("Plugin %s %sed", name, cmd));
+                    }
+                    catch (PluginNotFoundException e)
+                    {
+                        log.accept(format("Plugin %s not found", name));
+                    }
+                }
+                catch (ArgumentParserException e)
+                {
+                    log.accept("Error: " + e.getMessage());
+                    log.accept(ArgsParser.LOADER.formatHelp());
+                }
+            }
+            default ->
+            {
+                log.accept("Usage: /mcpm <install/uninstall/update/search/load/unload/reload/help>");
+            }
+        }
+    }
+}
--- a/src/main/java/org/hydev/mcpm/client/injector/PluginLoader.java
+++ b/src/main/java/org/hydev/mcpm/client/injector/PluginLoader.java
@@ -29,7 +29,7 @@ import static org.hydev.mcpm.utils.ReflectionUtils.setPrivateField;
 public class PluginLoader implements LoadBoundary, UnloadBoundary, ReloadBoundary
 {
     @Override
-    public void loadPlugin(String name) throws PluginNotFoundException
+    public boolean loadPlugin(String name) throws PluginNotFoundException
     {
         // 1. Find plugin file by name
         var nf = new PluginNotFoundException(name);
@@ -45,28 +45,30 @@ public class PluginLoader implements LoadBoundary, UnloadBoundary, ReloadBoundar
                 catch (IOException ignored) { return false; }
             }).findFirst().orElseThrow(() -> nf);
 
-        loadPlugin(file);
+        return loadPlugin(file);
     }
 
     @Override
-    public void loadPlugin(File jar)
+    public boolean loadPlugin(File jar)
     {
         // 2. Load plugin
         var pm = Bukkit.getPluginManager();
         try
         {
             var plugin = pm.loadPlugin(jar);
-            assert plugin != null;
+            if (plugin == null) return false;
 
             // 3. Call onLoad()
             plugin.onLoad();
             pm.enablePlugin(plugin);
+            return true;
         }
         catch (InvalidPluginException | InvalidDescriptionException e)
         {
             // These are errors indicating that the plugin we're trying to load is badly formatted.
             // There are nothing we can do.
             e.printStackTrace();
+            return false;
         }
     }
 
--- a/src/main/java/org/hydev/mcpm/client/injector/PluginLoader.java
+++ b/src/main/java/org/hydev/mcpm/client/injector/PluginLoader.java
@@ -21,7 +21,7 @@ import static org.hydev.mcpm.utils.ReflectionUtils.getPrivateField;
 import static org.hydev.mcpm.utils.ReflectionUtils.setPrivateField;
 
 /**
- * TODO: Write a description for this class!
+ * Implementation of plugin hot-loading/unloading
  *
  * @author Azalea (https://github.com/hykilpikonna)
  * @since 2022-09-27
--- a/src/main/java/org/hydev/mcpm/client/injector/PluginLoader.java
+++ b/src/main/java/org/hydev/mcpm/client/injector/PluginLoader.java
@@ -70,22 +70,63 @@ public class PluginLoader implements LoadBoundary, UnloadBoundary, ReloadBoundar
         }
     }
 
-    /**
-     * Dynamically unload a local plugin through JVM reflections and classloader hacks
-     *
-     * @param name Loaded plugin name
-     */
-    public void unloadPlugin(String name)
+    @Override
+    public void unloadPlugin(String name) throws PluginNotFoundException
     {
-        // TODO: Implement this
-        throw new UnsupportedOperationException("TODO");
+        var pm = Bukkit.getPluginManager();
+
+        // 1. Find plugin by name
+        var plugin = Arrays.stream(pm.getPlugins()).filter(p -> p.getName().equalsIgnoreCase(name)).findFirst()
+            .orElseThrow(() -> new PluginNotFoundException(name));
+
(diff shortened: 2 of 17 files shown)


src/main/java/org/hydev/mcpm/Controller.java
```diff
@@ -0,0 +1,80 @@
...
+import static org.hydev.mcpm.utils.Sugar.subFrom;
+
+
+/**
+ * Controller of the program
+ *
+ * @author Azalea (https://github.com/hykilpikonna)
+ * @since 2022-10-30
+ */
+public class Controller
```
1whatleytay, line 19: This doesn't work very nicely as a Controller for our whole operation IMO. Why does it call ArgsParser (Dependency Inversion Principle)? Why does the Controller know about the load/unload/reload commands that our CLI interface doesn't support? I'm suggesting:

 - Rename it to "PluginFacade" (or a name of your choice).
  - Make the interface concrete. Don't pass in raw arguments- not every interface that invokes our class should have to know/assemble/go through CLI arguments. I'm liking more and more the idea of a Controller that gets a list of Commands in its constructor but I don't think we're not at that point yet.
  - Move this to a server-only package so spigot dependency things are kept nicely away from all the "core" stuff.
  
Feel free to justify/talk more about the Controller class, but I think the actual Controller class should be pushed into PRs related to CommandInteractor.

src/main/java/org/hydev/mcpm/Controller.java
```diff
@@ -0,0 +1,80 @@
...
+     * Run a command
+     *
+     * @param args Command-line arguments (not including the command name/path)
+     * @param log Callback logger
+     */
+    public void runCommand(String[] args, Consumer<String> log)
+    {
+        if (args.length == 0) return;
+
+        var cmd = args[0].toLowerCase();
```
1whatleytay, line 43: Why not use subcommands? https://argparse4j.github.io/usage.html#sub-commands

src/main/java/org/hydev/mcpm/client/injector/PluginLoader.java
```diff
@@ -1,45 +1,135 @@
...
-     * @param jar Local jar file path
-     */
+    @Override
+    public void loadPlugin(String name) throws PluginNotFoundException
+    {
+        // 1. Find plugin file by name
+        var nf = new PluginNotFoundException(name);
+        var dir = new File("plugins");
+        if (!dir.isDirectory()) throw nf;
+        var file = Arrays.stream(Optional.ofNullable(dir.listFiles()).orElseThrow(() -> nf))
```
1whatleytay: This is fine, but note that a stack trace is assembled when the exception is constructed IIRC. It's better to create your exception on the line where you throw it so you have exact information about when/what went wrong.

Note how orElseThrow takes a lambda with no parameters.

src/main/java/org/hydev/mcpm/client/injector/PluginLoader.java
```diff
@@ -1,45 +1,135 @@
...
     public void loadPlugin(File jar)
     {
-        // TODO: Implement this
-        throw new UnsupportedOperationException("TODO");
+        // 2. Load plugin
+        var pm = Bukkit.getPluginManager();
+        try
+        {
+            var plugin = pm.loadPlugin(jar);
+            assert plugin != null;
```
1whatleytay: As a caller I'd probably want to know if loadPlugin failed (and I'd probably be able to say something nice to our user instead of panicking).

I think it's a good idea to make this return a discardable boolean or return some kind of failure indicator along with the assert (I'm talking about the catch stmt below too), but I understand if this is a rare case and you just want to leave it like this.

src/main/java/org/hydev/mcpm/client/injector/PluginLoader.java
```diff
@@ -1,45 +1,135 @@
...
+        var plugin = Arrays.stream(pm.getPlugins()).filter(p -> p.getName().equalsIgnoreCase(name)).findFirst()
+            .orElseThrow(() -> new PluginNotFoundException(name));
+
+        // 2. Unload plugin
+        pm.disablePlugin(plugin);
+
+        // 3. Unregister listeners
+        getPrivateField(pm, "listeners", new TypeToken<Map<Event, SortedSet<RegisteredListener>>>(){})
+            .ifPresent(listeners -> listeners.values().forEach(set -> set.removeIf(l -> l.getPlugin() == plugin)));
+
```
1whatleytay, line 90: Would it be possible to fork spigot to expose this? I'm not a fan of these kinds of sneaky maneuvers. If you bump your spigot version each one of these statements might suddenly break.

src/main/java/org/hydev/mcpm/client/injector/PluginLoader.java
```diff
@@ -1,45 +1,135 @@
...
+        // 6. Remove lookup name
+        getPrivateField(pm, "lookupNames", new TypeToken<Map<String, Plugin>>(){})
+            .ifPresent(names -> names.remove(plugin.getName()));
+
+        // 7. Unload Java classes so that the jar can be deleted on Windows
+        if (plugin.getClass().getClassLoader() instanceof URLClassLoader cl)
+        {
+            if (!setPrivateField(cl, "plugin", null) || !setPrivateField(cl, "pluginInit", null))
+                System.err.println("Error in setPrivateField, skipping unload");
```
1whatleytay, line 115: :flushed: After all that work?

src/main/java/org/hydev/mcpm/utils/Sugar.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.Arrays;
+
+/**
+ * Syntax sugar that java lacks
+ *
+ * @author Azalea (https://github.com/hykilpikonna)
+ * @since 2022-10-30
+ */
+public class Sugar
```
1whatleytay, line 11: Maybe a different name for this (NegativeSubscriptHelperFriendManFactoryBuilderConstructorPattern)? I was thinking about the cooking ingredient when I saw this lol. Alternatively, do you need the negative subscripting behaviour over copyOfRange? I don't see you using this class anywhere.

src/test/java/org/hydev/mcpm/utils/ReflectionUtilsTest.java
```diff
@@ -0,0 +1,53 @@
...
+    {
+        var a = new A();
+        var f = ReflectionUtils.getPrivateField(a, "field", new TypeToken<List<String>>(){}).orElseThrow();
+        assert f.get(0).equals("meow");
+        assert f.get(1).equals("qwq");
+
+        // The obtained field is a pointer
+        f.set(1, "wolf");
+        var newF = ReflectionUtils.getPrivateField(a, "field", new TypeToken<List<String>>(){}).orElseThrow();
+        assert newF.get(1).equals("wolf");
```
1whatleytay: You should try to shoot for 1 assert "group" per test method (here you assert before and after mutation). It sets a weird precedent where you encourage everyone to cram all your "xyzMethod" tests into this one "xyzMethod" test method. You do this also in SugarTest.java (and there you have some really good cases that you can separate out into many methods! like negative/positive subscripts).

src/main/java/org/hydev/mcpm/Controller.java
```diff
@@ -0,0 +1,80 @@
...
+                    {
+                        switch (cmd)
+                        {
+                            case "load" -> loader.loadPlugin(name);
+                            case "unload" -> unloader.unloadPlugin(name);
+                            case "reload" -> reloader.reloadPlugin(name);
+                        }
+                        log.accept(format("Plugin %s %sed", name, cmd));
+                    }
+                    catch (PluginNotFoundException e)
```
1whatleytay, line 63: Can't this be another catch to the parent try? `try { ... } catch (PluginNotFoundE) { ... } catch (ArgumentParserE) { ... }`

src/main/java/org/hydev/mcpm/Controller.java
```diff
@@ -0,0 +1,80 @@
...
+        reloader = pl;
+    }
+
+    /**
+     * Run a command
+     *
+     * @param args Command-line arguments (not including the command name/path)
+     * @param log Callback logger
+     */
+    public void runCommand(String[] args, Consumer<String> log)
```
1whatleytay, line 39: I love this use of reactive Consumer here.

1whatleytay (approved):

src/main/java/org/hydev/mcpm/client/injector/PluginLoader.java
hykilpikonna: But we are just a package manager, we kind of don't have the authority to ask Spigot to support us, or ask any server admin to use our own fork of Spigot. There are already many forks of Spigot available, like PaperMC, Tuinity, Glowstone, Purpur (which I personally use), and they all come with different levels of tweaks and optimizations. We can't ask any admin to use our unoptimized fork.

src/main/java/org/hydev/mcpm/client/injector/PluginLoader.java
hykilpikonna: This is just for unloading the classloader so that the jar file can be unlocked and deleted on Windows. Unix doesn't have this problem because Unix doesn't lock files.

src/main/java/org/hydev/mcpm/client/injector/PluginLoader.java
1whatleytay: We're not asking spigot to support us. By forking spigot we're opting into fixing changes that might affect our success at compile/merge time.

Are we trying to write code that supports all the forks that you've mentioned 😅? I feel like each fork could have different optimizations/layouts/values for the classes we're trying to manipulate.

src/main/java/org/hydev/mcpm/Controller.java
hykilpikonna: Not really? It can't use the `name` variable in the exception handling if it is another clause of the parent try statement

src/main/java/org/hydev/mcpm/Controller.java
hykilpikonna: Oops, I didn't notice that subcommands are supported 🥲 

We can change this when we add another command, or we can also keep it this way so that the help messages are shorter

src/main/java/org/hydev/mcpm/client/injector/PluginLoader.java
hykilpikonna: They actually don't. Their optimizations are mostly on in-game logic (it probably makes sense because plugin loading is normally only done once when the server starts, so there is not much need for optimizations)

In fact, I tested this on Purpur, and it works even when Purpur is a fork of Tuinity which is a fork of Paper which is a fork of Spigot which is a fork of CraftBukkit.

1whatleytay: We can keep using reflection then :eyes:

I think optimizing how plugins are loaded seems like a reasonable thing to expect a fork to do, whether it's for performance (I want to precompute these things to save time later!) or for utility of the server operator (let's make it easy for the operator to change things). I think you've made some really good points though, I think I'm just suggesting doing a lot of work for something that ultimately makes our tool *less* useful.

src/main/java/org/hydev/mcpm/utils/Sugar.java
hykilpikonna: Lol

I didn't end up using it but it's nice to have in our code

src/test/java/org/hydev/mcpm/utils/ReflectionUtilsTest.java
hykilpikonna: Okay I just made mutation into a separate test

src/main/java/org/hydev/mcpm/client/injector/PluginLoader.java
hykilpikonna: Fixed.

src/main/java/org/hydev/mcpm/Controller.java
hykilpikonna: Ok maybe we can ask the TA about how we should design the controller

PR #12, opened 2022-10-30
[+] Progress Bar

hykilpikonna:
1. Add ConsoleUtils for manipulating the Xterm console
2. Multiline simultaneous progress bar display
3. Speed and ETA calculation for the progress bar

https://user-images.githubusercontent.com/22280294/198893945-6cc6bf4e-2aa1-4db5-8882-c0744998f26b.mp4

### DIFF
--- a/src/main/java/org/hydev/mcpm/client/interaction/ProgressBar.java
+++ b/src/main/java/org/hydev/mcpm/client/interaction/ProgressBar.java
@@ -1,14 +1,28 @@
 package org.hydev.mcpm.client.interaction;
 
+import org.fusesource.jansi.AnsiConsole;
+import org.hydev.mcpm.utils.ConsoleUtils;
+
+import java.io.PrintStream;
+import java.util.*;
+
+import static java.lang.String.format;
+import static org.hydev.mcpm.utils.GeneralUtils.safeSleep;
+
 /**
- * TODO: Write a description for this class!
+ * Terminal progress bar based on Xterm escape codes
  *
  * @author Azalea (https://github.com/hykilpikonna)
  * @since 2022-09-27
  */
 public class ProgressBar implements AutoCloseable
 {
+    private final ConsoleUtils cu;
     private final ProgressBarTheme theme;
+    private final PrintStream out;
+    private final int cols;
+
+    private final List<ProgressRow> activeBars;
 
     /**
      * Create and initialize a progress bar
@@ -18,50 +32,44 @@ public class ProgressBar implements AutoCloseable
     public ProgressBar(ProgressBarTheme theme)
     {
         this.theme = theme;
-        this.init();
-    }
-
-    /**
-     * Initialize the progress bar (print the first line)
-     */
-    public void init()
-    {
-        // TODO: Implement this
-        throw new UnsupportedOperationException("TODO");
+        this.out = System.out;
+        this.cu = new ConsoleUtils(this.out);
+        this.activeBars = new ArrayList<>();
+        this.cols = AnsiConsole.getTerminalWidth();
+        System.out.println(cols);
     }
 
     /**
      * Append a progress bar at the end
      *
-     * @return Unique identifier of the progress bar
+     * @param bar Row of the progress bar
+     * @return bar for fluent access
      */
-    public String appendBar()
+    public ProgressRow appendBar(ProgressRow bar)
     {
-        // TODO: Implement this
-        throw new UnsupportedOperationException("TODO");
+        this.activeBars.add(bar);
+        out.println("");
+        update();
+        return bar;
     }
 
-    /**
-     * Set progress for a bar
-     *
-     * @param id Unique identifier
-     * @param progress Progress as a ratio in range 0-1
-     */
-    public void setBar(String id, float progress)
+    private void update()
     {
-        // TODO: Implement this
-        throw new UnsupportedOperationException("TODO");
+        // Roll back to the first line
+        cu.curUp(activeBars.size());
+        activeBars.forEach(bar -> out.println(bar.fmt(theme, cols)));
     }
 
     /**
      * Finish a progress bar
      *
-     * @param id Unique identifier of the progress bar
+     * @param bar Progress bar
      */
-    public void finishBar(String id)
+    public void finishBar(ProgressRow bar)
     {
-        // TODO: Implement this
-        throw new UnsupportedOperationException("TODO");
+        this.activeBars.remove(bar);
+        out.println();
+        update();
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/main/java/org/hydev/mcpm/client/interaction/ProgressBar.java
```diff
@@ -70,9 +97,36 @@ public void finishBar(String id)
...
+    public static void main(String[] args)
+    {
+        try (var b = new ProgressBar(ProgressBarTheme.ASCII_THEME))
+        {
+            //var r = b.appendBar(new ProgressRow(1000));
+            //for (int i = 0; i < 1000; i++)
+            //{
+            //    r.increase(1);
+            //    safeSleep(5);
+            //}
```
1whatleytay: Drop?

src/main/java/org/hydev/mcpm/client/interaction/ProgressBar.java
```diff
@@ -18,50 +34,61 @@ public class ProgressBar implements AutoCloseable
     public ProgressBar(ProgressBarTheme theme)
     {
         this.theme = theme;
-        this.init();
-    }
+        this.out = System.out;
+        this.cu = new ConsoleUtils(this.out);
+        this.activeBars = new ArrayList<>();
+        this.cols = AnsiConsole.getTerminalWidth();
```
1whatleytay, line 46: Gave this a test run on my local machine. I've had bad luck with accomplishing these kinds of things without going through ncurses, and it seems like what we have right now is broken with whatever IntelliJ configuration I have set up.

Seems like `AnsiConsole.getTerminalWidth()` is returning 0, which is causing ProgressRow#toString line 72 to pass a negative count to repeat.

Screenshot 2022-10-30 at 3 56 49 PM

I'll try to run outside of IntelliJ later :)

src/main/java/org/hydev/mcpm/client/interaction/ProgressBar.java
```diff
@@ -18,50 +34,61 @@ public class ProgressBar implements AutoCloseable
...
-    public void setBar(String id, float progress)
+    protected void update()
     {
-        // TODO: Implement this
-        throw new UnsupportedOperationException("TODO");
+        // Check time, update only every 0.0166s (60 fps)
+        // Performance of the update heavily depends on the terminal's escape code handling
+        // implementation, so frequent updates will degrade performance on a bad terminal
+        var curTime = System.nanoTime();
+        if ((curTime - lastUpdate) / 1e9d < 0.0166) return;
```
1whatleytay: Magic numbers? Make these constants or take them in as a parameter.

src/main/java/org/hydev/mcpm/client/interaction/ProgressRow.java
```diff
@@ -0,0 +1,134 @@
...
+    public String toString(ProgressBarTheme theme, int cols)
+    {
+        // Calculate speed. TODO: Use a moving window to calculate speed
+        double speed = completed / elapsed();
+        double eta = total / speed;
+        long etaS = (long) (eta % 60), etaM = (long) (eta / 60);
+
+        // Replace variables
+        var p = format("%3.0f%%", 100d * completed / total);
+        var t = fmt.replace("{prefix}", theme.prefix())
```
1whatleytay, line 60: Not trying to say we should pull in a whole templating library just for this, but can we do anything to sidestep String#replace? Weird edge cases include theme.suffix() returning something like `"{eta}"` or `desc` being `"{progbar} {progbar} {progbar}"`.

1whatleytay (dismissed):

src/main/java/org/hydev/mcpm/client/interaction/ProgressBar.java
hykilpikonna: Yea IntelliJ uses Gradle to run the project, and Gradle's output isn't a tty, and Xterm control sequences only support tty environments. So you can only test it using external terminals or manually ran using IntelliJ's built-in terminal outside of gradle. I've included instructions in the readme

src/main/java/org/hydev/mcpm/client/interaction/ProgressBar.java
hykilpikonna: It's written in the comment above, that 0.0166 is the delay between each frame for 60fps. And `1e9` is pretty straightforward for converting nanoseconds to seconds.

hykilpikonna: I guess I can make fps configurable

1whatleytay: This is actually an anti-pattern IIRC. Comments help but it's wayy better to give them a formal name and move them out the code "battle zone."

(Again I'm taking this class to learn but I've personally been called out for this kinda stuff before :P)

src/main/java/org/hydev/mcpm/client/interaction/ProgressRow.java
hykilpikonna: Yea but we can just make sure that doesn't happen since we're only using it internally in our program, and it's really unlikely to happen just by chance...

src/main/java/org/hydev/mcpm/client/interaction/ProgressBar.java
1whatleytay: Thanks! Can you detect this so if the TA is tinkering they don't hit an exception 7 levels deep? DW if you're busy with other things, but if it's a 2 second fix I'd recommend tackling it!

hykilpikonna: Okay just fixed this

src/main/java/org/hydev/mcpm/client/interaction/ProgressBar.java
hykilpikonna: Yea dropped

src/main/java/org/hydev/mcpm/client/interaction/ProgressRow.java
1whatleytay: What about file names? Those are user controlled so someone who knows about our utility can break this class. Again, unlikely, but it's spooky to give the user control over our format strings.

hykilpikonna: This is why the description is replaced at the very last so that the placeholders in the description wouldn't be replaced by anything else

src/main/java/org/hydev/mcpm/client/interaction/ProgressBar.java
```diff
@@ -18,61 +40,118 @@ public class ProgressBar implements AutoCloseable
...
-     * @param progress Progress as a ratio in range 0-1
-     */
-    public void setBar(String id, float progress)
+    protected void update()
+    {
+        // Check time to limit for framerate (default 60fps)
+        // Performance of the update heavily depends on the terminal's escape code handling
+        // implementation, so frequent updates will degrade performance on a bad terminal
+        var curTime = System.nanoTime();
+        if ((curTime - lastUpdate) / 1e9d < frameDelay) return;
```
1whatleytay, line 84: I think 1e9d is still a bit of a magic number (S_TO_PS or something). You're call if you want to factor it out.

1whatleytay (approved):

thudoan1706 (approved):

jerryzhu509 (approved):

PR #13, opened 2022-11-01
GET /db

1whatleytay:
I refused to go along with the `feat-` branch prefix but I'll fold if anyone calls me out on it.

This PR adds:
 - CreateDatabase class which is in charge of generating a database object from a SpigetCrawler directory (this is NOT configurable, it must go through the filesystem).
 - It also defines the layout of the /db endpoint.

It does NOT:
 - Define how this method is run OR tie it into an existing flow for the SpigetCrawler (this needs to be run). I'm willing to add a flow if necessary.


1whatleytay:
Pretty sure my brace style is inconsistent with the rest of the project. Do you mind if we set the linter to go with inline braces? This seems to be the accepted style for Java projects.

hykilpikonna:
> Pretty sure my brace style is inconsistent with the rest of the project. Do you mind if we set the linter to go with inline braces? This seems to be the accepted style for Java projects.

Inline or next-line brace placement is one of the least concerning for code style, as it doesn't influence readability either way. I personally prefer next-line placement because I feel like it makes code more readable when lines are more spaced out, and it also makes the look of the code more symmetric, which aligns more with people's aesthetics. I've seem many java projects using next-line braces in the past, and I always used next-line braces in any C-family languages like java, typescript, Kotlin, etc. I think we don't need to set a standardized brace placement rule for our project.

src/main/java/org/hydev/mcpm/server/crawlers/SpigetCrawler.java
```diff
@@ -44,7 +44,7 @@
     public SpigetCrawler(File dataDir)
     {
         this.dataDir = dataDir;
-        this.blacklist = new StoredHashMap<>(new File(dataDir, "crawler/spiget/blacklist.json"));
+        this.blacklist = new StoredHashMap<>(new File(dataDir, "crawler/spiget/blacklist.gs"));
```
hykilpikonna: Why was the file name changed? 🤔

hykilpikonna (dismissed):

src/main/java/org/hydev/mcpm/server/crawlers/spiget/CreateDatabase.java
```diff
@@ -0,0 +1,174 @@
...
+     * @param databaseFile Path to the db.json file that will be written to.
+     */
+    public static void writeDatabase(File crawlerDirectory, File databaseFile) {
+        Database database = createDatabase(crawlerDirectory);
+
+        if (database == null)
+            return;
+
+        try {
+            new ObjectMapper().writeValue(databaseFile, database);
```
hykilpikonna, line 63: It would be nice if you can implement zstd compression here as well. You can try https://github.com/luben/zstd-jni

1whatleytay: Yeah I agree. Do you want me to do this In-PR? I'm tempted to push this off to when DatabaseInteractor has some meat to it (mostly because I'm lazy TBH).

src/main/java/org/hydev/mcpm/server/crawlers/SpigetCrawler.java
1whatleytay: I have absolutely no clue LOL. Fixed.

1whatleytay: I even checked the diff before committing. I was wondering why this was showing up in the status dialog.

hykilpikonna (dismissed):

src/main/java/org/hydev/mcpm/server/crawlers/spiget/CreateDatabase.java
hykilpikonna: Okay yea that's reasonable

src/main/java/org/hydev/mcpm/server/crawlers/spiget/CreateDatabase.java
```diff
@@ -0,0 +1,174 @@
...
+            .map(Optional::get)
+            .toList();
+
+        return new Database(plugins);
+    }
+
+    @NotNull
+    private static Optional<PluginModel> createPluginModel(File directory) {
+        try {
+            var id = Long.parseLong(directory.getName());
```
MstrPikachu, line 123: I see the exception being caught but are directory names just integers

1whatleytay: The spiget API that Azalea writes outputs directories with names corresponding to their plugin id (which are just integers, yep). We want to be prudent, so when we're fetching the plugin id we want to be able to handle cases where some directory is a rebel and has a non-integer name (we just skip it, it's probably not a plugin).

MstrPikachu (dismissed):
Good stuff

hykilpikonna (dismissed):

src/main/java/org/hydev/mcpm/client/models/PluginYml.java
```diff
@@ -18,46 +19,25 @@
...
-    }
-
-    /**
-     * YAML parsing requires an no-args constructor
-     */
-    public PluginYml()
-    {
-        this("", "", "");
-    }
+    public static class InvalidPluginMetaStructure extends Exception { }
```
hykilpikonna, line 41: What a long name lol

src/main/java/org/hydev/mcpm/server/crawlers/spiget/CreateDatabase.java
```diff
@@ -0,0 +1,186 @@
...
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
+
+            System.out.println("Failed to write database file.");
+        } catch (Exception e) {
+            e.printStackTrace();
+
+            System.out.println("Failure in DB creation.");
+        }
```
hykilpikonna: Hmm, ArchLinux actually has "lastupdate" and "lastsync" endpoints which give unix timestamps. I think it's because they have a huge mirror list and they need an automatic system to check which ones are maintained and which ones are outdated (since just having the hash doesn't tell you about update time, just tells you whether it is outdated).

So if you want, we can write a timestamp file along with the hash file. Should be an easy change.

image

image

hykilpikonna (dismissed):

src/main/java/org/hydev/mcpm/server/crawlers/spiget/CreateDatabase.java
1whatleytay: Yeah I can do that :D 

What are we doing with the timestamp though? Does it matter if we have an outdated by 2 months file or an outdated by 1 month file?

hykilpikonna: Maybe (if we have demand in the future) we can make an automatic mirror tracker that tracks mirrors' availability and update frequency, and calculates score for each mirror like https://archlinux.org/mirrors/status/.

src/main/java/org/hydev/mcpm/client/models/PluginYml.java
1whatleytay: Unknown

PR #15, opened 2022-11-02
[+] MCPRS Plugin Repository Server Mirrors

hykilpikonna:
1. Created a docker image for automatic mirror creation
2. Wrote a LOT of documentation
3. Created 2 mirrors, one in Luxembourg and one in Tokyo
4. Created MirrorSelector to automatically select the fastest mirror
5. Created NetworkUtils to ping
6. Tested MirrorSelector and NetworkUtils
7. Added a python script `tools/run` to build and run any class externally

### DIFF
--- a/src/main/java/org/hydev/mcpm/client/Launcher.java
+++ b/src/main/java/org/hydev/mcpm/client/Launcher.java
@@ -10,5 +10,6 @@ public class Launcher
 {
     public static void main(String[] args)
     {
+        System.out.println("Meow");
     }
 }
--- a/src/main/java/org/hydev/mcpm/client/database/Mirror.java
+++ b/src/main/java/org/hydev/mcpm/client/database/Mirror.java
@@ -1,0 +1,27 @@
+package org.hydev.mcpm.client.database;
+
+import java.util.List;
+
+/**
+ * Model for a mirror source
+ *
+ * @param host Hosting domain or half url of the mirror (e.g. "mcprs.hydev.org" or "eg.com/mcprs")
+ * @param protocols Supported protocols. Can contain "http", "https", or "rsyncd"
+ * @param tier Mirror tier. 0th tier mirrors are the source. 1st tier mirrors are synced from 0th tier mirrors, etc.
+ * @param country Origin of the mirror (e.g. Canada)
+ * @param speed Actual upstream speed in mbps, can be tested on speedtest.net
+ * @param interval Interval in hours between updates
+ *
+ * @author Azalea (https://github.com/hykilpikonna)
+ * @since 2022-11-01
+ */
+public record Mirror(
+    String host,
+    List<String> protocols,
+    int tier,
+    String country,
+    int speed,
+    int interval
+)
+{
+}
--- a/src/main/java/org/hydev/mcpm/client/database/Mirror.java
+++ b/src/main/java/org/hydev/mcpm/client/database/Mirror.java
@@ -1,5 +1,6 @@
 package org.hydev.mcpm.client.database;
 
+import javax.annotation.Nullable;
 import java.util.List;
 
 /**
@@ -24,4 +25,25 @@ public record Mirror(
     int interval
 )
 {
+    /**
+     * Check if the mirror supoorts either http or https.
+     *
+     * @return Supports web
+     */
+    public boolean isWeb()
+    {
+        return protocols.contains("http") || protocols.contains("https");
+    }
+
+    /**
+     * Get the url of the mirror. Set to use https in priority.
+     *
+     * @return Http URL or null if it doesn't support http protocols
+     */
+    public @Nullable String url()
+    {
+        if (protocols.contains("https")) return "https://" + host;
+        else if (protocols.contains("http")) return "http://" + host;
+        else return null;
+    }
 }
--- a/src/main/java/org/hydev/mcpm/client/database/Mirror.java
+++ b/src/main/java/org/hydev/mcpm/client/database/Mirror.java
@@ -1,17 +1,18 @@
 package org.hydev.mcpm.client.database;
 
 import javax.annotation.Nullable;
-import java.util.List;
+import java.util.Set;
 
 /**
  * Model for a mirror source
  *
- * @param host Hosting domain or half url of the mirror (e.g. "mcprs.hydev.org" or "eg.com/mcprs")
+ * @param host Hosting domain of the mirror (e.g. "mcprs.hydev.org")
  * @param protocols Supported protocols. Can contain "http", "https", or "rsyncd"
  * @param tier Mirror tier. 0th tier mirrors are the source. 1st tier mirrors are synced from 0th tier mirrors, etc.
  * @param country Origin of the mirror (e.g. Canada)
  * @param speed Actual upstream speed in mbps, can be tested on speedtest.net
  * @param interval Interval in hours between updates
+ * @param httpEndpoint Endpoint to the http file server (e.g. "abc" means "mcprs.hydev.org/abc")
  *
  * @author Azalea (https://github.com/hykilpikonna)
  * @since 2022-11-01
@@ -22,7 +23,10 @@ public record Mirror(
     int tier,
     String country,
     int speed,
-    int interval
+    int interval,
+
+    // TODO: Add defaults value when parsing for httpEndpoint
+    String httpEndpoint
 )
 {
     /**
@@ -42,8 +46,8 @@ public record Mirror(
      */
     public @Nullable String url()
     {
-        if (protocols.contains("https")) return "https://" + host;
-        else if (protocols.contains("http")) return "http://" + host;
+        if (protocols.contains("https")) return "https://" + host + httpEndpoint();
+        else if (protocols.contains("http")) return "http://" + host + httpEndpoint();
         else return null;
     }
 
@@ -51,5 +55,14 @@ public record Mirror(
     public String toString()
     {
         return host;
+    }
+
+    @Override
+    public String httpEndpoint()
+    {
+        if (httpEndpoint == null) return "/";
+
+        // Ensure leading /
+        return httpEndpoint.startsWith("/") ? httpEndpoint : "/" + httpEndpoint;
     }
 }
--- a/src/main/java/org/hydev/mcpm/client/database/Mirror.java
+++ b/src/main/java/org/hydev/mcpm/client/database/Mirror.java
@@ -18,7 +18,7 @@ import java.util.List;
  */
 public record Mirror(
     String host,
-    List<String> protocols,
+    Set<String> protocols,
     int tier,
     String country,
     int speed,
--- a/src/main/java/org/hydev/mcpm/client/database/Mirror.java
+++ b/src/main/java/org/hydev/mcpm/client/database/Mirror.java
@@ -46,4 +46,10 @@ public record Mirror(
         else if (protocols.contains("http")) return "http://" + host;
         else return null;
     }
+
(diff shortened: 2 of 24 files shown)


thudoan1706 (approved):

thudoan1706 (dismissed):

src/main/java/org/hydev/mcpm/client/Launcher.java
```diff
@@ -10,5 +10,6 @@
 {
     public static void main(String[] args)
     {
+        System.out.println("Meow");
```
MstrPikachu, line 13: Meow?

src/main/java/org/hydev/mcpm/client/database/Mirror.java
```diff
@@ -0,0 +1,55 @@
...
+ * @param country Origin of the mirror (e.g. Canada)
+ * @param speed Actual upstream speed in mbps, can be tested on speedtest.net
+ * @param interval Interval in hours between updates
+ *
+ * @author Azalea (https://github.com/hykilpikonna)
+ * @since 2022-11-01
+ */
+public record Mirror(
+    String host,
+    List<String> protocols,
```
1whatleytay: Unless it screws with serialization:

```suggestion
    Set<String> protocols,
```

src/main/java/org/hydev/mcpm/client/database/Mirror.java
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Get the url of the mirror. Set to use https in priority.
+     *
+     * @return Http URL or null if it doesn't support http protocols
+     */
+    public @Nullable String url()
+    {
+        if (protocols.contains("https")) return "https://" + host;
+        else if (protocols.contains("http")) return "http://" + host;
+        else return null;
```
1whatleytay: Maybe it would be better to have a `List<URI>` for each Mirror... I don't mind repeating strings in a config file.

MstrPikachu (dismissed):
Java is good

src/main/java/org/hydev/mcpm/client/database/MirrorSelector.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Implementation of mirror select boundary
+ *
+ * @author Azalea (https://github.com/hykilpikonna)
+ * @since 2022-11-01
+ */
+public class MirrorSelector implements MirrorSelectBoundary
+{
+    public static final String MIRROR_LIST_URL = "https://mcprs.hydev.org/mirrorlist.yml";
```
1whatleytay, line 22: lol

src/main/java/org/hydev/mcpm/client/database/MirrorSelector.java
```diff
@@ -0,0 +1,64 @@
...
+        // Try parsing the YML
+        YML.readValue(yml, new TypeReference<List<Mirror>>(){});
+
+        // Save yml to local folder
+        LOCAL_PATH.getParentFile().mkdirs();
+        Files.writeString(LOCAL_PATH.toPath(), yml);
+    }
+
+    @Override
+    public List<Pair<Mirror, Integer>> pingMirrors() throws IOException
```
1whatleytay:
```suggestion
    public static int pingMirror(Mirror mirror) throws IOException
```

Is pinging a responsibility of MirrorSelector? Can any parent class do this with `listAvailableMirrors`? This is probably safe to remove from the Boundary.

src/main/java/org/hydev/mcpm/client/interaction/ProgressBar.java
```diff
@@ -56,8 +55,8 @@ public ProgressBar(ProgressBarTheme theme)
         this.frameDelay = 1 / 60d;
 
         // Check if output is a TTY. If not, change frame rate to 0.5 fps to avoid spamming a log.
-        this.istty = isatty(STDOUT_FILENO) == 0;
-        if (istty) this.frameDelay = 1 / 0.5;
+        this.istty = isatty(1) == 1;
```
1whatleytay: I getcha, `#define STDOUT_FILENO 1`. Any reason you removed the reference to the global constant? It makes it more clear what you're trying to do.

src/main/java/org/hydev/mcpm/utils/NetworkUtils.java
```diff
@@ -0,0 +1,62 @@
...
+{
+    private static final Timeout ONE_S = Timeout.ofSeconds(1);
+    private static final int PING_ITERS = 3;
+
+    static
+    {
+        // Preheat HTTP request module so that subsequent access return similar delay
+        try
+        {
+            Request.head("https://1.1.1.1").connectTimeout(ONE_S).execute();
```
1whatleytay: Does this have side effects or does it cost a bit? It might be better to shift this onto the first iteration of the ping method if possible (e.g. `lazy static var req = Request.head(...).execute()`, in ping: `discard(req)`).

src/main/java/org/hydev/mcpm/utils/NetworkUtils.java
```diff
@@ -0,0 +1,62 @@
...
+
+        var start = System.currentTimeMillis();
+
+        // Do request and check code
+        var hasSuccess = IntStream.range(0, PING_ITERS).anyMatch(i ->
+        {
+            try
+            {
+                return Request.head(u).connectTimeout(ONE_S).responseTimeout(ONE_S)
+                    .execute().returnResponse().getCode() < 400;
```
1whatleytay, line 52: Nit: I had a double take on this haha. Do you think it would help to give this expression a name `var requestSucceeded = httpResponseCode < 400;`?

src/main/java/org/hydev/mcpm/utils/Pair.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
+
+/**
+ * Simple unmodifiable generic pair
+ *
+ * @author Azalea (https://github.com/hykilpikonna)
+ * @since 2022-11-01
+ */
+public record Pair<K, V>(K k, V v) implements Map.Entry<K, V>
```
1whatleytay, line 13: I am flat out floored this isn't provided in Java. Spooky stuff.

1whatleytay (dismissed):

src/main/java/org/hydev/mcpm/client/Launcher.java
hykilpikonna: 🐱

src/main/java/org/hydev/mcpm/client/database/Mirror.java
hykilpikonna: Maybe we should, but it also means that we have to type/paste the URL multiple times which increases the chance for a typo

1whatleytay: Perhaps, but it *might* be worth the flexibility IMO. EX One mirror could have a subdomain just for rsync (not sure what this actually gives us). Your call, if you don't want to do this feel free to resolve.

src/main/java/org/hydev/mcpm/client/database/MirrorSelector.java
hykilpikonna: Pinging an url is not, which is why the function for pinging an URL is placed in NetworkUtils. 

I don't think having 

```java
static int pingMirror(Mirror mirror)
```

is meaningful, because it's just the equivalent of calling `ping()` on the mirror's URL. One-line functions aren't really helpful.

src/main/java/org/hydev/mcpm/client/database/Mirror.java
hykilpikonna: Okay yea I think we could separate out the rsync URL with the HTTP url

src/main/java/org/hydev/mcpm/client/database/MirrorSelector.java
1whatleytay: Was thinking something similar last night, but that's not the core point I want to bring up. I'm more confused why this method is a part of the Boundary. Why do you have to implement a ping method if your job is supposed to be to give a list of Mirrors?

hykilpikonna: Because it's supposed to be a mirror selector, like automatically selecting the fastest mirror. The `pingMirrors` function returns a sorted list, so people can just choose the first item to be the fastest mirror, or ask the user to select from them.

src/main/java/org/hydev/mcpm/utils/NetworkUtils.java
hykilpikonna: It costs like ~500ms when the class is loaded for the first time, which does the same thing as you are suggesting.

src/main/java/org/hydev/mcpm/utils/NetworkUtils.java
hykilpikonna: Lol I'll add a comment

src/main/java/org/hydev/mcpm/client/database/MirrorSelector.java
1whatleytay: No, that's not what I mean. https://github.com/CSC207-2022F-UofT/mcpm/pull/15/files#diff-d236b867dff295147af99249ea5c8531470c9c91860efa983bffcfdf60274182R35 implies that different `MirrorSelector`s will have different implementations of the `pingMirrors` method. Is that what you're going for? I can't imagine another implementation or why implementation could vary for different MirrorSelectorBoundary subclasses.

src/main/java/org/hydev/mcpm/utils/NetworkUtils.java
1whatleytay: I guess it's okay if the ping method is the only way to interact with NetworkUtils. It kinda sounds like a general class that might have different methods on the future, so forcing 500ms for the first time interact with any component of the class is a bit spooky. I'll leave this up to you whether or not you want to change it. EX PingUtils can't go wrong.

src/main/java/org/hydev/mcpm/client/interaction/ProgressBar.java
hykilpikonna: Yea I was testing if changing it to STDIN_FILENO (0) would have any difference, but there isn't a constant for STDIN_FILENO so I wrote a number and forgot to change it back.

src/main/java/org/hydev/mcpm/utils/NetworkUtils.java
hykilpikonna: Okay actually I'll initialize it on the first ping call

src/main/java/org/hydev/mcpm/client/database/MirrorSelector.java
hykilpikonna: Ok then should we maybe change this into an abstract / interface default method? Or should we just remove it?

1whatleytay: I think it could do it's job as a static method somewhere! It could even still be an instance method to MirrorSelector I just think this method implementation is fixed and every class that implements MirrorSelectorBoundary shouldn't need to copy paste this snippet.

hykilpikonna: Okay I just made it into a default interface method

hykilpikonna:
Merging because it previously had 3 approving reviews, and no subsequent changes are significant enough to request new reviews.

PR #18, opened 2022-11-03
Implement database syncing and List Fetching

1whatleytay:
This PR implements some basic database syncing methods and the list all plugins boundary.

There's one bug with the progress bar implementation that causes the progress to jump to full after the request is finished. I'll fix that tonight :)

### DIFF
--- a/src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
+++ b/src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
@@ -1,10 +1,7 @@
 package org.hydev.mcpm.client.database.fetcher;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import org.apache.hc.client5.http.classic.methods.HttpGet;
 import org.apache.hc.client5.http.fluent.Request;
-import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
-import org.apache.hc.client5.http.impl.classic.HttpClients;
 import org.apache.hc.core5.http.ClassicHttpResponse;
 import org.apache.hc.core5.http.HttpEntity;
 import org.hydev.mcpm.client.models.Database;
@@ -14,13 +11,15 @@ import org.jetbrains.annotations.Nullable;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.net.URI;
-import java.net.http.HttpClient;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.security.NoSuchAlgorithmException;
 
+/**
+ * A database fetcher that has a local persistent cache along and a network fallback.
+ */
 public class LocalDatabaseFetcher implements DatabaseFetcher {
     private final URI host;
     private final Path cacheDirectory;
@@ -32,10 +31,24 @@ public class LocalDatabaseFetcher implements DatabaseFetcher {
 
     public static final String USER_AGENT = "MCPM Client";
 
+    /**
+     * Creates a database fetcher that syncs with a mirror.
+     * The cache directory will default to .mcpm
+     *
+     * @param host The host URL MCPM server mirror that will be used to request database files.
+     *             Example: "http://mcpm.hydev.com"
+     */
     public LocalDatabaseFetcher(URI host) {
         this(host, Path.of(".mcpm/"));
     }
 
+    /**
+     * Creates a database fetcher that syncs with a mirror.
+     *
+     * @param host The host URL MCPM server mirror that will be used to request database files.
+     *             Example: "http://mcpm.hydev.com"
+     * @param cacheDirectory The directory that the cached database and hash file will be stored (as child files).
+     */
     public LocalDatabaseFetcher(URI host, Path cacheDirectory) {
         this.host = host;
         this.cacheDirectory = cacheDirectory;
@@ -45,7 +58,7 @@ public class LocalDatabaseFetcher implements DatabaseFetcher {
         return Request
             .get(URI.create(host.toString() + "/" + path))
             .addHeader("Host", host.getHost())
-            .addHeader("User-Agent", "MCPM Client")
+            .addHeader("User-Agent", USER_AGENT)
             .addHeader("Accepts", "application/json");
     }
 
--- a/src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
+++ b/src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
@@ -76,7 +76,7 @@ public class LocalDatabaseFetcher implements DatabaseFetcher {
     @Nullable
     private String fetchLocalHash() {
         try {
-            return Files.readString(cacheDirectory.resolve(DATABASE_FILE_NAME));
+            return Files.readString(Paths.get(cacheDirectory.toString(), HASH_FILE_NAME));
         } catch (IOException e) {
             return null;
         }
@@ -104,7 +104,7 @@ public class LocalDatabaseFetcher implements DatabaseFetcher {
                 return localDatabase;
             }
 
-            var file = cacheDirectory.resolve(DATABASE_FILE_NAME).toFile();
+            var file = Paths.get(cacheDirectory.toString(), DATABASE_FILE_NAME).toFile();
 
             var database = new ObjectMapper().readValue(file, Database.class);
 
--- a/src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
+++ b/src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
@@ -1,0 +1,179 @@
+package org.hydev.mcpm.client.database.fetcher;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.hc.client5.http.classic.methods.HttpGet;
+import org.apache.hc.client5.http.fluent.Request;
+import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
+import org.apache.hc.client5.http.impl.classic.HttpClients;
+import org.apache.hc.core5.http.ClassicHttpResponse;
+import org.apache.hc.core5.http.HttpEntity;
+import org.hydev.mcpm.client.models.Database;
+import org.hydev.mcpm.utils.HashUtils;
+import org.jetbrains.annotations.Nullable;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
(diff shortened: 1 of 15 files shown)


hykilpikonna:
This is great! I'll review it in a moment.

In the future, can we split a feature into many smaller detailed commits rather than one commit that changes everything? For example, moving the hash function to a HashUtil class could be its own commit.

1whatleytay:
Yeah! I can do that. This is actually a squashed commit lol 😅 

There are still some issues I'm cleaning up. I'm not going to mark it as WIP since there aren't any regressions though. I don't think I'm going to merge compression into this PR either.

src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
```diff
@@ -0,0 +1,192 @@
...
+import java.io.IOException;
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.security.NoSuchAlgorithmException;
+
+/**
+ * A database fetcher that has a local persistent cache along and a network fallback.
```
hykilpikonna, line 21: Please add your `@author` and `@since` in javadocs

Or you can add a file template header or something in IDEA

Mine:

```
/**
 * TODO: Write a description for this class!
 *
 * @author Hykilpikonna (https://github.com/hykilpikonna)
 * @since ${YEAR}-${MONTH}-${DAY}
 */
```

src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
```diff
@@ -0,0 +1,192 @@
...
+
+    /**
+     * Creates a database fetcher that syncs with a mirror.
+     * The cache directory will default to .mcpm
+     *
+     * @param host The host URL MCPM server mirror that will be used to request database files.
+     *             Example: "http://mcpm.hydev.com"
+     */
+    public LocalDatabaseFetcher(URI host) {
+        this(host, Path.of(".mcpm/"));
```
hykilpikonna, line 42: Hmm... should this be in a local folder? Or should it be in the user home directory like how I did in MirrorSelector?

For one part, I think it would be nice if we don't need to sync database again running the program in different locations, but for the other part, do you think home directory is an assumption we're making that some servers might not have?

src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
```diff
@@ -0,0 +1,192 @@
...
+            return response.returnContent().asString().trim();
+        } catch (IOException e) {
+            return null;
+        }
+    }
+
+    @Nullable
+    private String fetchLocalHash() {
+        try {
+            return Files.readString(cacheDirectory.resolve(DATABASE_FILE_NAME));
```
hykilpikonna: 🤔

Did you forget to calculate the hash?

src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
```diff
@@ -0,0 +1,192 @@
...
+    @Nullable
+    private Database fetchLocalDatabase() {
+        try {
+            if (localDatabase != null) {
+                return localDatabase;
+            }
+
+            var file = cacheDirectory.resolve(DATABASE_FILE_NAME).toFile();
+
+            var database = new ObjectMapper().readValue(file, Database.class);
```
hykilpikonna, line 109: Can you use the `JACKSON` constant instead of creating a new ObjectMapper every time?

src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
```diff
@@ -0,0 +1,192 @@
...
+        }
+    }
+
+    private boolean localDatabaseIsUpToDate() {
+        var hostHash = fetchHostHash();
+        var localHash = fetchLocalHash();
+
+        // We can't reach the server right now, don't update.
+        if (hostHash == null)
+            return true;
```
hykilpikonna, line 91: Maybe throw a custom exception here so that client can tell the user what's going on

src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
hykilpikonna: Or should this be HASH_FILE_NAME?

hykilpikonna: Would it be better to calculate hash on the fly instead?

src/main/java/org/hydev/mcpm/client/database/inputs/ListPackagesInput.java
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Input for the ListAllPackagesBoundary.
+ *
+ * @param itemsPerPage The amount of items for each page.
+ *                     If negative, pageNumber is ignored and all plugins are returned.
+ * @param pageNumber The page number (starting at 0).
+ * @param noCache If true, the ListPackagesBoundary will re-download the database before performing the request.
+ */
+public record ListPackagesInput(
```
hykilpikonna, line 11: Maybe add a default no-args constructor, because not all arguments are going to be used.

hykilpikonna (dismissed):

src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
1whatleytay: This should be `HASH_FILE_NAME` yep, this is already fixed for some reason on local :eyes:. The database file is huge and I expect reading + calculating its hash is going to be expensive. Plus, the local hash is set up to lean on the side of integrity (the user probably won't see incorrect behaviour from an out of sync hash file).

src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
1whatleytay: I'm not a fan of big Constants files, I feel like it creates an unnecessary dependency and an extra barrier for reusing this class elsewhere. Plus if you need to pull the Constants file for one specific class, you end up pulling a ton of unnecessary "Constant" stuff from around the project along with it. I can change it to use Constants.java if you really want, but I would rather create its own constant in-class if creating an `ObjectMapper` is expensive or keep this inline if it's not.

I've moved out some of the properties of the Constants.java mappers into the targeting classes anyway, so (I don't think?) there should be a change in behaviour... Point me out if I'm wrong.

src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
1whatleytay: It doesn't matter. There's no restriction on where the cache directory needs to be. There's a constructor below this that allows you to specify where you want to put your cache directory. You can treat this constructor as a demo initializer if you'd like, I know this directory is supposed to be used for the scraper.

src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
1whatleytay: Actually, I'd like to keep this as a silent failure. This `return true` doesn't abort anything, it just uses the on-disk database if it's there. I think that would enable users to search/list/check for updates on packages when our server is down or our internet isn't quite working (e.g. UOFT CSC207 Tutorial Internet).

src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
hykilpikonna: Ok yea that makes sense

src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
hykilpikonna: Oh yea right

src/main/java/org/hydev/mcpm/client/database/inputs/ListPackagesInput.java
1whatleytay: This is an interesting thing to point out. What exactly is the purpose of the list all packages command? I don't think it makes sense for it to be strictly "listing all packages" because the parent has to have access to a DatabaseFetcher object, and listing all packages with that object is as simple as `fetcher.database().plugins()`.

hykilpikonna (approved):

jerryzhu509:
src/main/java/org/hydev/mcpm/client/database/DatabaseInteractor.java

Why delete the search functions, boundaries, and the input class? I thought DatabaseInteractor also implements the search methods, which I made before.

jerryzhu509:
Also, remember to add java doc on methods. Some in-line comments here and there would also be very helpful

1whatleytay:
We have a linter that requires all methods/classes to be javadoc'd, theoretically I should have everything commented. If there's a spot I missed, can you point it out in a PR review?

PS. I'm not sure about the search related stuff. They seem empty though, so I think it's best to fill them in a PR related to adding search functionality.

jerryzhu509:
Okay then maybe it's fine. I was just looking through ProgressBarFetcherListener.java and LocalDatabaseFetcher.java and some private methods weren't commented. 

Yeah makes sense to delete those. I can make a PR for the setup for the search use case for you guys to quickly review and get into implementation once the listing functionality here goes through.

jerryzhu509 (approved):

src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
```diff
@@ -0,0 +1,192 @@
...
+            return null;
+        }
+    }
+
+    @Override
+    @Nullable
+    public Database fetchDatabase(boolean cache, DatabaseFetcherListener listener) {
+        var databaseHasAlreadyBeenChecked = localDatabase != null;
+
+        if (cache && (databaseHasAlreadyBeenChecked || localDatabaseIsUpToDate())) {
```
MstrPikachu, line 183: What if the local database has already been checked but it's out of date? Then it would still return the local version.

1whatleytay: Yeah, I thought of this but this was intentionally left like this due to two assumptions:
 - Instances of the CLI will be short lived (we will only check our database once per instance).
 - If we have a long living instance (that handles multiple commands, for example), users can set the `cache` parameter to false to **force** a redownload of the database.

The body of this if statement is expensive, so I want to minimize the need to make multiple subsequent network requests if multiple commands run `fetchDatabase` in sequence. TLDR the user can opt-out of this `databaseHasAlreadyBeenChecked` variable so I think it's okay.

MstrPikachu (approved):

PR #21, opened 2022-11-14
Structure of Search Use Case

jerryzhu509:
Made some classes and an interface implemented in DatabaseInteractor. I'd appreciate your reviews on whether the setup is okay. The rest is to implement the search methods. (11/13/2022)

EDIT (11/20/2022):
This PR was originally for setting up the structure for the search use case. I created the SearchBoundary, SearchPackagesInput, SearchPackagesResult. Because of the difficulty of getting things merged (needed 3 approvals before), I ended up just doing almost of my work in this one PR. I'll separate them into the rough stages:
1. Set up the structure (original PR) and implemented two methods searchByName and searchByKeyword.
2.  Implemented methods searchByName and searchByKeyword in the DatabaseInteractor and did if statements in there originally.
3. Realized that the design was not great. Thanks to a suggestion from the TA, better is to apply the factory design pattern to create Searchers. To this end, added Searcher interface, the SearchBy... classes that implemented the interact, and a SearchFactory that created searchers based on the input type.

### DIFF
--- a/src/main/java/org/hydev/mcpm/client/database/DatabaseInteractor.java
+++ b/src/main/java/org/hydev/mcpm/client/database/DatabaseInteractor.java
@@ -112,19 +112,23 @@ public class DatabaseInteractor implements ListPackagesBoundary, SearchPackagesB
             return SearchPackagesResult.by(SearchPackagesResult.State.INVALID_INPUT);
 
         var plugins = database.plugins();
-        if (input.type() == 1) {
+        if (input.type() == SearchPackagesInput.Type.BY_NAME) {
             return new SearchPackagesResult(
                     SearchPackagesResult.State.SUCCESS,
                     searchByName(plugins, searchStr));
         }
-        else if (input.type() == 2) {
+        else if (input.type() == SearchPackagesInput.Type.BY_COMMAND) {
             return new SearchPackagesResult(
                     SearchPackagesResult.State.SUCCESS,
                     searchByCommand(plugins, searchStr));
         }
-        return new SearchPackagesResult(
-                SearchPackagesResult.State.SUCCESS,
-                searchByKeyword(plugins, searchStr));
+        else if (input.type() == SearchPackagesInput.Type.BY_KEYWORD) {
+            return new SearchPackagesResult(
+                    SearchPackagesResult.State.SUCCESS,
+                    searchByKeyword(plugins, searchStr));
+        } else {
+            return SearchPackagesResult.by(SearchPackagesResult.State.INVALID_SEARCH_TYPE);
+        }
     }
 
     @Override
--- a/src/main/java/org/hydev/mcpm/client/database/DatabaseInteractor.java
+++ b/src/main/java/org/hydev/mcpm/client/database/DatabaseInteractor.java
@@ -1,22 +1,27 @@
 package org.hydev.mcpm.client.database;
 
 import org.hydev.mcpm.client.database.boundary.ListPackagesBoundary;
+import org.hydev.mcpm.client.database.boundary.SearchPackagesBoundary;
 import org.hydev.mcpm.client.database.fetcher.DatabaseFetcher;
 import org.hydev.mcpm.client.database.fetcher.DatabaseFetcherListener;
 import org.hydev.mcpm.client.database.fetcher.LocalDatabaseFetcher;
 import org.hydev.mcpm.client.database.fetcher.ProgressBarFetcherListener;
 import org.hydev.mcpm.client.database.inputs.ListPackagesInput;
-import org.hydev.mcpm.client.database.inputs.ListPackagesResult;
+import org.hydev.mcpm.client.database.results.ListPackagesResult;
+import org.hydev.mcpm.client.database.inputs.SearchPackagesInput;
+import org.hydev.mcpm.client.database.results.SearchPackagesResult;
+import org.hydev.mcpm.client.models.PluginModel;
 
 import java.net.URI;
 import java.util.ArrayList;
+import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
 /**
  * Handles fetching and performing operations on the plugin database.
  */
-public class DatabaseInteractor implements ListPackagesBoundary {
+public class DatabaseInteractor implements ListPackagesBoundary, SearchPackagesBoundary {
     private final DatabaseFetcher fetcher;
     private final DatabaseFetcherListener listener;
 
@@ -34,7 +39,7 @@ public class DatabaseInteractor implements ListPackagesBoundary {
      * Creates a new database with the provided fetcher and upload listener.
      *
      * @param fetcher The fetcher that will be used to request the database object in boundary calls.
-     * @param listener The listener that will receives updates if the database is downloaded from the internet.
+     * @param listener The listener that will receive updates if the database is downloaded from the internet.
      */
     public DatabaseInteractor(DatabaseFetcher fetcher, DatabaseFetcherListener listener) {
         this.fetcher = fetcher;
@@ -86,6 +91,65 @@ public class DatabaseInteractor implements ListPackagesBoundary {
             result,
             database.plugins().size()
         );
+    }
+
+    /**
+     * Searches for plugins based on the provided name, keyword, or command.
+     * The input contains the type of search.
+     *
+     * @param input Record of inputs as provided in SearchPackagesInput. See it for more info.
+     * @return Packages result. See the SearchPackagesResult record for more info.
+     */
+    public SearchPackagesResult search(SearchPackagesInput input) {
+        var database = fetcher.fetchDatabase(!input.noCache(), listener);
+
+        if (database == null) {
+            return SearchPackagesResult.by(SearchPackagesResult.State.FAILED_TO_FETCH_DATABASE);
+        }
+
+        var searchStr = input.searchStr();
+        if (searchStr.isEmpty())
+            return SearchPackagesResult.by(SearchPackagesResult.State.INVALID_INPUT);
+
+        var plugins = database.plugins();
+        if (input.type() == 1) {
+            return new SearchPackagesResult(
+                    SearchPackagesResult.State.SUCCESS,
+                    searchByName(plugins, searchStr));
+        }
+        else if (input.type() == 2) {
+            return new SearchPackagesResult(
+                    SearchPackagesResult.State.SUCCESS,
+                    searchByCommand(plugins, searchStr));
+        }
+        return new SearchPackagesResult(
+                SearchPackagesResult.State.SUCCESS,
+                searchByKeyword(plugins, searchStr));
+    }
+
+    @Override
+    public List<PluginModel> searchByName(List<PluginModel> plugins, String name)
+    {
+//        for (PluginModel plugin : plugins) {
+//            // Assuming name is the same in every version.
+//            System.out.println(plugin.versions().meta().name());
+//        }
+        // Todo
+        return null;
+    }
+
+    @Override
+    public List<PluginModel> searchByKeyword(List<PluginModel> plugins, String keyword)
+    {
+        // Todo
+        return List.of();
+    }
+
+    @Override
+    public List<PluginModel> searchByCommand(List<PluginModel> plugins, String command)
+    {
+        // Todo
+        return List.of();
     }
 
     /**
--- a/src/main/java/org/hydev/mcpm/client/database/DatabaseInteractor.java
+++ b/src/main/java/org/hydev/mcpm/client/database/DatabaseInteractor.java
@@ -12,7 +12,7 @@ import java.util.Map;
  * @author Azalea (https://github.com/hykilpikonna)
  * @since 2022-09-27
  */
-public class DatabaseInteractor
+public class DatabaseInteractor implements SearchPluginBoundary
 {
     /** Name index: map[lower-cased name] = Plugin of that name */
     private final Map<String, PluginModel> nameIndex;
@@ -35,15 +35,12 @@ public class DatabaseInteractor
         this.commandIndex = new HashMap<>();
 
         // TODO: Implement this
+
+
         throw new UnsupportedOperationException("TODO");
     }
 
-    /**
-     * Find the plugin by name
-     *
-     * @param name Name of the plugin
-     * @return Plugin of that name, or null if not found
-     */
+    @Override
     public PluginModel findByName(String name)
     {
         if (nameIndex.containsKey(name)) {
@@ -52,19 +49,7 @@ public class DatabaseInteractor
         return null;
     }
 
-    /**
-     * Search for a plugin by keyword.
-     * <p>
-     * This function should ignore letter case iin searching.
-     * For example, searching "java" would match "Java" as well
-     * <p>
-     * This function implements fuzzy search that doesn't require the exact phrase to be available,
-     * but requires all words in the phrase to be present.
-     * For example, searching "java 11" would match "java jdk 11" but not "java"
-     *
-     * @param keyword Keyword
-     * @return List of packages matching the keyword, or empty list
-     */
+    @Override
     public List<PluginModel> searchByKeyword(String keyword)
     {
         if (keywordIndex.containsKey(keyword)) {
@@ -73,12 +58,7 @@ public class DatabaseInteractor
         return List.of();
     }
 
-    /**
-     * Search for a plugin by command or command alias.
-     *
-     * @param command Command name
-     * @return Plugins that provides the command, or empty list
-     */
+    @Override
     public List<PluginModel> searchByCommand(String command)
     {
         if (commandIndex.containsKey(command)) {
--- a/src/main/java/org/hydev/mcpm/client/database/DatabaseInteractor.java
+++ b/src/main/java/org/hydev/mcpm/client/database/DatabaseInteractor.java
@@ -101,6 +101,7 @@ public class DatabaseInteractor implements ListPackagesBoundary, SearchPackagesB
      * @param input Record of inputs as provided in SearchPackagesInput. See it for more info.
      * @return Packages result. See the SearchPackagesResult record for more info.
      */
+    @Override
     public SearchPackagesResult search(SearchPackagesInput input) {
         var database = fetcher.fetchDatabase(!input.noCache(), listener);
 
--- a/src/main/java/org/hydev/mcpm/client/database/DatabaseInteractor.java
+++ b/src/main/java/org/hydev/mcpm/client/database/DatabaseInteractor.java
@@ -11,7 +11,6 @@ import org.hydev.mcpm.client.database.results.ListPackagesResult;
 import org.hydev.mcpm.client.database.inputs.SearchPackagesInput;
 import org.hydev.mcpm.client.database.results.SearchPackagesResult;
 import org.hydev.mcpm.client.database.searchusecase.SearcherFactory;
-import org.hydev.mcpm.client.models.PluginModel;
 
 import java.net.URI;
 import java.util.*;
@@ -19,16 +18,13 @@ import java.util.stream.Collectors;
 
 /**
  * Handles fetching and performing operations on the plugin database.
+ *
+ * @author Taylor Whatley
+ * @author Jerry Zhu (<a href="https://github.com/jerryzhu509">...</a>)
  */
 public class DatabaseInteractor implements ListPackagesBoundary, SearchPackagesBoundary {
     private final DatabaseFetcher fetcher;
     private final DatabaseFetcherListener listener;
-
-    /**
-     * Map associating each search type to the map associating a plugin feature to its
-     * corresponding list of plugins.
-     */
-    private final Map<SearchPackagesInput.Type, Map<String, List<PluginModel>>> searchMaps = new HashMap<>();
 
     /**
      * Creates a new database with the provided database fetcher.
@@ -49,9 +45,6 @@ public class DatabaseInteractor implements ListPackagesBoundary, SearchPackagesB
     public DatabaseInteractor(DatabaseFetcher fetcher, DatabaseFetcherListener listener) {
         this.fetcher = fetcher;
         this.listener = listener;
-        for (SearchPackagesInput.Type type: SearchPackagesInput.Type.values()) {
-            this.searchMaps.put(type, null);
-        }
     }
 
     /**
@@ -115,19 +108,14 @@ public class DatabaseInteractor implements ListPackagesBoundary, SearchPackagesB
             return SearchPackagesResult.by(SearchPackagesResult.State.FAILED_TO_FETCH_DATABASE);
         }
 
-        var searchStr = input.searchStr();
+        var searchStr = input.searchStr().toLowerCase();
         if (searchStr.isEmpty())
             return SearchPackagesResult.by(SearchPackagesResult.State.INVALID_INPUT);
 
         var plugins = database.plugins();
 
-        // Make map if null
-        SearcherFactory factory = new SearcherFactory();
-        if (this.searchMaps.get(input.type()) == null)
-            this.searchMaps.put(input.type(), factory.createSearcher(input).constructSearchMaps(plugins));
-
         return new SearchPackagesResult(SearchPackagesResult.State.SUCCESS,
-                this.searchMaps.get(input.type()).get(searchStr));
+                SearcherFactory.createSearcher(input).getSearchList(searchStr, plugins));
     }
 
     /**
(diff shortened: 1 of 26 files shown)


src/main/java/org/hydev/mcpm/client/database/inputs/SearchPackagesInput.java
```diff
@@ -0,0 +1,13 @@
...
+ * Input for the SearchPackagesBoundary.
+ *
+ * @param type An int that specifies the type of search.
+ *             1 if the search is by name, 2 if the search is by command. Else the search is by keyword.
+ * @param searchStr String containing the name, keyword, or command to search by.
+ * @param noCache If true, the ListPackagesBoundary will re-download the database before performing the request.
+ *
+ * @author Jerry Zhu (https://github.com/jerryzhu509)
+ */
+public record SearchPackagesInput(int type, String searchStr, boolean noCache) {}
```
1whatleytay: `type` is opaque. What happens if I pass in `0`, `-1` or `700`? Please use an `enum`.

1whatleytay (dismissed):

src/main/java/org/hydev/mcpm/client/database/inputs/SearchPackagesInput.java
jerryzhu509: great point. using enum instead now

src/main/java/org/hydev/mcpm/client/database/DatabaseInteractor.java
```diff
@@ -88,6 +93,69 @@ public ListPackagesResult list(ListPackagesInput input) {
...
+            return new SearchPackagesResult(
+                    SearchPackagesResult.State.SUCCESS,
+                    searchByCommand(plugins, searchStr));
+        }
+        else if (input.type() == SearchPackagesInput.Type.BY_KEYWORD) {
+            return new SearchPackagesResult(
+                    SearchPackagesResult.State.SUCCESS,
+                    searchByKeyword(plugins, searchStr));
+        } else {
+            return SearchPackagesResult.by(SearchPackagesResult.State.INVALID_SEARCH_TYPE);
```
1whatleytay: Nice! Bonus suggestion: If you put this in a valued switch case you don't need to worry about having an invalid search type, there will be a compile error if your switch isn't exhaustive (I think):

```java
switch (input.type()) {
  case SearchPackagesInput.Type.BY_NAME -> { ... }
  case SearchPackagesInput.Type.BY_COMMAND -> { ... }
  case SearchPackagesInput.Type.BY_KEYWORD -> { ... }
}
```

1whatleytay (dismissed):

src/main/java/org/hydev/mcpm/client/database/DatabaseInteractor.java
jerryzhu509: switch is definitely cleaner. thx. what should it return by default though?

src/main/java/org/hydev/mcpm/client/database/DatabaseInteractor.java
```diff
@@ -88,6 +101,35 @@ public ListPackagesResult list(ListPackagesInput input) {
...
+        var searchStr = input.searchStr();
+        if (searchStr.isEmpty())
+            return SearchPackagesResult.by(SearchPackagesResult.State.INVALID_INPUT);
+
+        var plugins = database.plugins();
+
+        // Make map if null
+        SearcherFactory factory = new SearcherFactory();
+        if (this.searchMaps.get(input.type()) == null)
+            this.searchMaps.put(input.type(), factory.createSearcher(input).constructSearchMaps(plugins));
```
MstrPikachu: Here, you create a factory even if your index for that use case is populated.
But you avoid creating a factory by createSearcher a static method.

Although it seems to me that the searcher objects themselves should be able to have the search method as static, static methods defined in an interface cannot be overriden which wouldn't work.

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherFactory.java
```diff
@@ -0,0 +1,30 @@
...
+ * @author Jerry Zhu (<a href="https://github.com/jerryzhu509">...</a>)
+ */
+public class SearcherFactory {
+
+    /**
+     * Returns the new searcher object based on the input type.
+     *
+     * @param input Contains the search type in particular. See SearchPackagesInput for details.
+     */
+    public Searcher createSearcher(SearchPackagesInput input) {
```
1whatleytay: `private SearcherFactory() { ... }`

```suggestion
    public static Searcher createSearcher(SearchPackagesInput input) {
```

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByName.java
```diff
@@ -0,0 +1,36 @@
...
+     * Not case-sensitive.
+     * @param plugins A list of all plugins in the database.
+     * @return A dictionary associating the name to the matching plugins.
+     */
+    @Override
+    public Map<String, List<PluginModel>> constructSearchMaps(List<PluginModel> plugins) {
+        Map<String, List<PluginModel>> models = new HashMap<>();
+        for (PluginModel plugin : plugins) {
+            // Get latest version
+            var v = plugin.versions().stream().max(Comparator.comparingLong(PluginVersion::id));
```
1whatleytay: This should be a method of PluginModel. It's duplicated among all your search classes.

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByCommand.java
```diff
@@ -0,0 +1,41 @@
...
+     * @return A dictionary associating the command to the matching plugins.
+     */
+    @Override
+    public Map<String, List<PluginModel>> constructSearchMaps(List<PluginModel> plugins) {
+        Map<String, List<PluginModel>> models = new HashMap<>();
+        for (PluginModel plugin : plugins) {
+            // Get latest version
+            var v = plugin.versions().stream().max(Comparator.comparingLong(PluginVersion::id));
+            if (v.isPresent()) {
+                String[] keywords = v.get().meta().description().toLowerCase().split(" ");
```
1whatleytay: The description has commands?

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByKeyword.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public class SearcherByKeyword implements Searcher {
+
+    /**
+     * Returns a dictionary mapping the different keywords to the matching plugins
+     * This function should ignore letter case in searching.
+     * For example, searching "java" would match "Java" as well.
+     * This function implements fuzzy search that doesn't require the exact phrase to be available,
+     * but requires all words in the phrase to be present.
+     * For example, searching "java 11" would match "java jdk 11" but not "java"
```
MstrPikachu, line 22: The memory complexity required for fuzzy searching would be infinite given the way the code is currently set up. 
At the moment, you pass the index all the way to the controller.
However, fuzzy searching requires extra work to be done during runtime, and it cannot be precomputed (you can still precompute things but you need to do extra work). 
I suggest that instead of having your searchers return an index, they take in a the parameters and return the search result. This way you can implement fuzzy searching in the inner layer instead of the outer layer, which is not supposed to deal with these things.

src/main/java/org/hydev/mcpm/client/database/searchusecase/Searcher.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+public interface Searcher {
+
+    /**
+     * Searches for plugins based on the provided user input.
+     *
+     * @param plugins A list of all plugins in the database.
+     * @return A dictionary associating a string feature of the plugins to the matching plugins.
+     */
+    Map<String, List<PluginModel>> constructSearchMaps(List<PluginModel> plugins);
```
1whatleytay, line 21: This is great!

src/main/java/org/hydev/mcpm/client/database/DatabaseInteractor.java
```diff
@@ -98,15 +140,26 @@ public static void main(String[] args) {
...
+//        var text = result
+//            .plugins()
+//            .stream()
+//            .map(x -> x.versions().stream().findFirst().map(value -> value.meta().name()))
+//            .filter(Optional::isPresent)
+//            .map(Optional::get)
+//            .map(value -> "  " + value)
+//            .collect(Collectors.joining("\n"));
+//
+//        System.out.println(text);
```
1whatleytay: Uncomment or delete.

1whatleytay (dismissed):
LGTM

src/main/java/org/hydev/mcpm/client/database/DatabaseInteractor.java
jerryzhu509: ah im trying to debug rn. sth weird is going on

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByCommand.java
jerryzhu509: just copied and pasted search by keywords bc commands aren't yet implemented

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByName.java
jerryzhu509: dont see it yet.. maybe its in a PR?

1whatleytay: I'm asking you to add it lol 😛

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByKeyword.java
MstrPikachu: More precisely, every word in the search query must be in the plugin. This is the same as set intersection. Ideally we'd use sets for set intersection, but your interface doesn't allow for that. You could modify the interface to return the search result. The searcher objects can store their own index in a static variable

src/main/java/org/hydev/mcpm/client/database/DatabaseInteractor.java
jerryzhu509: createSearcher being static makes sense, and it's what i had. But a brief glance online had some ppl doing the factory design pattern without static. ig ill go with static since it makes more sense.

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByKeyword.java
jerryzhu509: I don't fully understand. Wdm by "pass the index all the way to the controller".
Searching by keywords just basically returns the plugins whose descriptions contain all the words within the "keyword" string, the essence of which is the keywords to list of plugins map. This is created in the business rules layer as far as i can tell

MstrPikachu: Right now, you create a mapping from single keywords to plugins.
However, this map is given to the database interactor.
The database interactor now has to do some logical work to determine which plugins have *all* the keywords in the search query, unless your map can take in every possible query and output a plugin for that.
The point is that all the work should be done in the innermost part in clean architecture, and the database interactor shouldn't have to worry about it, but this should be the job of the searcher.

jerryzhu509: Yeah the point of the searchers is to do the hard work in computing the plugins for *every* keyword. All the database interactor does in searching is a dictionary look up basically

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByName.java
jerryzhu509: ah lol sure ill do that once  i get stuff to work :(

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByKeyword.java
MstrPikachu: It only does every singular keyword. Currently it cannot do multiple keywords.

Suppose I did wanted to add a function to search plugins that have every keyword in my query.

One way I could do this, through clean architecture, is the following:

1. Add a new search type to the enum and create the corresponding searcher that implements the Searcher interface.
2. Implement the interface: I need to return a map. This needs to be a mapping from multiple keywords to a list of plugins. I don't really want to change the Searcher interface, which requires a mapping from String to list, but that's okay, we can represent multiple keywords as a string separated with spaces. No problem yet.
3. Now I need to go through every plugin, and see what possible combination of keywords could lead me to this plugin. Thankfully, if the search query contains a keyword not in my plugin's keywords, this plugin shouldn't be included. So we only consider keywords in the plugin.
4. Now I just need to add this plugin to every possible combination of keywords. Well, if I have N keywords, then every possible combination would be 2^N - 1 (we exclude the empty set). However, I only counted the subsets. Subsets are not ordered, but input queries can be in any order! This means I need to take every subset and also add the plugin to every permutation of it, so I can cover every query.
5. We went through every plugin and can return the map.

Now let's look at what happens above, and try to change parts that look like it took too much work to do something.

1. Step 1 is ok.
2. I could possibly change the interface to take in a set of strings.
3. This is ok.
4. Hm, input queries are in any order. If I had the query with me, I could sort all the keywords. This would solve my issue of subsets being unordered. Or, if I had the query with me, I could actually just use a mapping from single keywords to a list of plugins: The answer would be the intersection of the lists of plugins matching each keyword. Ideally I would also change the mapping to use sets, so I can speed up the intersection.
5. In the previous step I just computed the answer, so we can return the answer instead.

This is the fundamental problem: most the hard work is done by the searcher. All the database interactor does is look up the answer. But what if we get a new problem where we can't look up the answer (unless we sacrifice something)? Then we have to move all the work to the searcher, the searcher can give the answer to the database interactor, and then for the specific case where you need to do a little extra, you can add it to the lower level class instead of the higher level class.

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByKeyword.java
```diff
@@ -0,0 +1,77 @@
...
+        if (SearcherByKeyword.keywordMap == null) {
+            SearcherByKeyword.keywordMap = constructSearchMaps(plugins);
+        }
+
+        String [] keywords = ((String) inp).toLowerCase().split(" "); // Should be a string
+        Set<PluginModel> res = new HashSet<>();
+        boolean first = true;
+        for (String keyword : keywords) {
+            List<PluginModel> pl = SearcherByKeyword.keywordMap.get(keyword);
+            if (first) {
```
MstrPikachu: Good job implementing it!
I noticed a few simplifications we can do. 
You noticed there is no universal set, so the first iteration has to add instead of intersect. 
But observe that doing this on the first iteration only is the same as doing this before the loop. This removes the if statement. 
But if you remove the addAll and replace it with an assignment, you’ll end up modifying the set, which modifies the index. However instead of using addAll on a new set you can use the set constructor directly to copy all the elements from the indexing set. 
Also the string cast is sus

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByKeyword.java
```diff
@@ -0,0 +1,77 @@
...
+    @Override
+    public Map<String, List<PluginModel>> constructSearchMaps(List<PluginModel> plugins) {
+        Map<String, List<PluginModel>> models = new HashMap<>();
+        for (PluginModel plugin : plugins) {
+            // Get latest version
+            var v = plugin.getLatestPluginVersion();
+            if (v.isPresent() && v.get().meta() != null &&
+                    v.get().meta().description() != null && !v.get().meta().description().equals("")) {
+                String[] keywords = v.get().meta().description()
+                        .replaceAll("[^a-zA-Z ]", "").toLowerCase().split("\\s+");
```
1whatleytay, line 36: Nice regex!

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByKeyword.java
```diff
@@ -0,0 +1,77 @@
...
+                for (String keyword : keywords) {
+                    if (!models.containsKey(keyword))
+                        models.put(keyword, new ArrayList<>());
+                    models.get(keyword).add(plugin);
```
1whatleytay, line 40: It might save some work to use `getOrDefault` here...
```suggestion
                    var keywordList = models.getOrDefault(keyword, null);
                    if (keywordList == null) {
                        keywordList = new ArrayList<>();
                        models.put(keyword, keywordList);
                    }
                    keywordList.add(plugin);
```

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByKeyword.java
```diff
@@ -0,0 +1,77 @@
...
+     */
+    @Override
+    public List<PluginModel> getSearchList(Object inp, List<PluginModel> plugins) {
+
+        // Instantiate if null
+        if (SearcherByKeyword.keywordMap == null) {
+            SearcherByKeyword.keywordMap = constructSearchMaps(plugins);
+        }
+
+        String [] keywords = ((String) inp).toLowerCase().split(" "); // Should be a string
```
1whatleytay: No type check here?
```suggestion
        if (!(inp instanceof String input)) { return null; } // guard let
        String [] keywords = input.toLowerCase().split(" "); // Should be a string
```

1whatleytay (dismissed):

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByKeyword.java
MstrPikachu: I didn’t see the carat so I thought it was deleting everything 💀
But isn’t the regex also deleting spaces? How do you split afterwards
Am I lacking critical information

1whatleytay: Honestly me too I have no idea what this matches to. Seems like it's targeting punctuation. `^` probably means exclusion or smth.
Screenshot 2022-11-16 at 7 53 12 PM

MstrPikachu: ^ means to exclude everything within the character class, in this case, every non (alphabetical character or space)

1whatleytay: Yeah so... the space is in the group right? We don't delete it :D

MstrPikachu: I have played myself
In that case, the split can just be by spaces `" +"` instead of by repeated whitespace `"\\s+"`

1whatleytay: What about newlines?

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByCommand.java
```diff
@@ -0,0 +1,40 @@
...
+    private static Map<String, List<PluginModel>> commandMap = null;
+
+    /**
+     * Returns a dictionary mapping the different commands to the matching plugins.
+     *
+     * @param plugins A list of all plugins in the database.
+     * @return A dictionary associating the command to the matching plugins.
+     */
+    @Override
+    public Map<String, List<PluginModel>> constructSearchMaps(List<PluginModel> plugins) {
```
MstrPikachu, line 24: Personally I would consider this an implementation detail, as we can see it already varies, for example the SearcherByKeyword actually uses sets of plugins in the map.
This means it shouldn't be exposed to the public (you can remove this from the interface).

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByCommand.java
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Searches for plugins based on the provided user input.
+     *
+     * @param inp User input for the search. Should be a name as a string here.
+     * @param plugins A list of all plugins in the database.
+     * @return A dictionary associating a string feature of the plugins to the matching plugins.
+     */
+    @Override
+    public List<PluginModel> getSearchList(Object inp, List<PluginModel> plugins) {
```
MstrPikachu, line 36: This is a default method, but I think a default behaviour for searching does not makes too much sense. You could leave this abstract instead, so the compiler will remind you to override the search method (which is a core functionality).

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByKeyword.java
MstrPikachu: Newlines would get deleted in the replaceAll... a bug?

src/main/java/org/hydev/mcpm/client/database/searchusecase/Searcher.java
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * Searches for plugins based on the provided user input.
+     *
+     * @param inp User input for the search.
+     * @param plugins A list of all plugins in the database.
+     * @return A dictionary associating a string feature of the plugins to the matching plugins.
+     */
+    List<PluginModel> getSearchList(Object inp, List<PluginModel> plugins);
```
MstrPikachu, line 30: You should use a SearchInput object instead of regular object

MstrPikachu (dismissed):
Good just fix the sus Object argument in the Searcher interface

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByCommand.java
1whatleytay: I disagree unless I'm misunderstanding what you're saying... Are you saying we want our Searcher interface to provide a default behaviour for this method?

MstrPikachu: I am saying we do not. I missed a word

1whatleytay: Oh good. How about `throw NotImplementedException()`?

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByKeyword.java
jerryzhu509: i prefer to keep what i had since it's shorter

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByKeyword.java
jerryzhu509: will do

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByKeyword.java
1whatleytay: Alright, just figure that you're doubling the work (searching the map twice, once in `containsKey` and once again in `get`).

src/main/java/org/hydev/mcpm/client/database/searchusecase/Searcher.java
jerryzhu509: Keywords should be on the same line. even if different lines i think we'd want to remove the new line?

jerryzhu509: ill actually have to adjust searchInput later.. the input isn't necessarily a string due to search by commands

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByKeyword.java
jerryzhu509: ah true getOrDefault is cleaner. ill do that

1whatleytay: its just a bit quicker lol

src/main/java/org/hydev/mcpm/client/database/searchusecase/SearcherByCommand.java
jerryzhu509: i dont follow exactly.  ill leave it for now

1whatleytay (approved):

MstrPikachu (approved):

PR #24, opened 2022-11-14
Add CheckForUpdates boundary

1whatleytay:
I've added the work around the CheckForUpdates boundary.

I've made sure to try to allow it to handle multiple types of inputs (by id/name/string) since I'm not sure what the lock file will contain (or what will be feasible).

I've also added a MatchPlugins boundary to allow other parts of the program to look up plugins based on information from a lock file if needed (hopefully).


1whatleytay:
Imma close this PR and wait until Search Packages is done. I think CheckForUpdates depends on some kind of Search functionality and there are some subtle problems that need to be solved that are out of scope.

PR #26, opened 2022-11-14
Downloader with pb

thudoan1706:
When reading a file, I only read 1024 bytes for each loop. I think it affects the speed of the progress bar and makes it slow.

### DIFF
--- a/src/main/java/org/hydev/mcpm/client/Downloader.java
+++ b/src/main/java/org/hydev/mcpm/client/Downloader.java
@@ -1,7 +1,18 @@
 package org.hydev.mcpm.client;
 
-import java.io.File;
+import org.hydev.mcpm.client.interaction.ProgressBar;
+import org.hydev.mcpm.client.interaction.ProgressBarTheme;
+import org.hydev.mcpm.client.interaction.ProgressRow;
+
+import java.io.*;
+import java.net.HttpURLConnection;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
+import java.net.URL;
+
+import static java.lang.String.format;
+import static org.hydev.mcpm.utils.GeneralUtils.safeSleep;
 
 /**
  * File downloader
@@ -17,16 +28,49 @@ public class Downloader
     /** Number of simultaneous downloads */
     private int threads = 5;
 
+
     /**
      * Download one file from the internet to local storage through HTTP request
      *
      * @param url Remote file URL
      * @param to Local file path
      */
+
     public void downloadFile(String url, File to)
     {
         // TODO: Implement this
-        throw new UnsupportedOperationException("TODO");
+        // throw new UnsupportedOperationException("TODO");
+        try {
+            URL link = new URL(url);
+            HttpURLConnection http = (HttpURLConnection)link.openConnection();
+            long fileSize = (long)http.getContentLengthLong();
+
+
+            BufferedInputStream in = new BufferedInputStream(http.getInputStream());
+            FileOutputStream fileos = new FileOutputStream(to);
+            BufferedOutputStream bout = new BufferedOutputStream(fileos, 1024);
+            byte[] buffer = new byte[1024];
+            int read = 0;
+
+            var b = new ProgressBar(ProgressBarTheme.ASCII_THEME);
+            var all = new ArrayList<ProgressRow>();
+            var row = new ProgressRow(fileSize)
+                    .unit("MB")
+                    .desc(format("Downloaded", all.size()))
+                    .descLen(30);
+            all.add(b.appendBar(row));
+            while ((read = in.read(buffer, 0, 1024)) >= 0) {
+
+                bout.write(buffer, 0, read);
+                all.forEach(a -> a.increase(1024));
+            }
+            bout.close();
+            in.close();
+            System.out.println("Download completed");
+        }
+        catch (IOException ex) {
+            ex.printStackTrace();
+        }
     }
 
     /**
@@ -40,8 +84,45 @@ public class Downloader
      */
     public void downloadFiles(Map<String, File> urls, boolean progress, int threads)
     {
-        // TODO: Implement this
-        throw new UnsupportedOperationException("TODO");
+        Thread[] threadss = new Thread[threads];
+        final int UrlPerThread = urls.size() / threads;
+        final int remainingUrl = urls.size() % threads;
+
+        for (int t = 0; t < threads; t++) {
+            final int thread = t;
+            threadss[t] = new Thread() {
+                @Override
+                public void run() {
+                    runThread(urls, threads, thread, UrlPerThread, remainingUrl);
+                }
+            };
+        }
+
+        for (Thread t: threadss) {
+            t.start();
+        }
+    }
+    private void runThread(Map<String, File> urls, int threads, int thread, int UrlPerThread, int remainingUrl) {
+        // Store all urls into a list
+        List<String> files = new ArrayList<>();
+        for (Map.Entry<String, File> url:urls.entrySet()) {
+            files.add(url.getKey());
+        }
+        // Store urls per thread
+        List<String> inFiles = new ArrayList<>();
+
+        for (int i=thread*UrlPerThread; i < (thread + 1) * UrlPerThread; i++) {
+            inFiles.add(files.get(i));
+        }
+        if (thread == threads-1 && remainingUrl > 0) {
+            for (int j = files.size() - remainingUrl; j < files.size(); j++) {
+                inFiles.add(files.get(j));
+            }
+        }
+
+        for (String url:inFiles) {
+            downloadFile(url, urls.get(url));
+            }
     }
 
     /**
--- a/src/main/java/org/hydev/mcpm/client/Downloader.java
+++ b/src/main/java/org/hydev/mcpm/client/Downloader.java
@@ -12,7 +12,6 @@ import java.util.Map;
 import java.net.URL;
 
 import static java.lang.String.format;
-import static org.hydev.mcpm.utils.GeneralUtils.safeSleep;
 
 /**
  * File downloader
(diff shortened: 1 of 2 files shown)


thudoan1706:
I just realize the logic for the progress bar is a bit odd since they are all independent for every file downloaded. so each progress bar only has one row

src/main/java/org/hydev/mcpm/client/Downloader.java
```diff
@@ -40,8 +83,45 @@ public void downloadFile(String url, File to)
...
+        for (Thread t: threadss) {
+            t.start();
+        }
+    }
+    private void runThread(Map<String, File> urls, int threads, int thread, int UrlPerThread, int remainingUrl) {
+        // Store all urls into a list
+        List<String> files = new ArrayList<>();
+        for (Map.Entry<String, File> url:urls.entrySet()) {
+            files.add(url.getKey());
+        }
```
1whatleytay, line 109: Consider `Set<String> files = urls.keySet()` or if you want a list:

```suggestion
        var files = urls.keySet().stream().toList();
```

src/main/java/org/hydev/mcpm/client/Launcher.java
```diff
@@ -7,9 +7,4 @@
  * @since 2022-09-27
  */
 public class Launcher
-{
-    public static void main(String[] args)
-    {
-        System.out.println("Meow");
-    }
-}
+{ public static void main(String[] args) {System.out.println("Meow");}}
```
1whatleytay, line 10: What happened here? 😅

src/main/java/org/hydev/mcpm/client/Downloader.java
```diff
@@ -40,8 +83,45 @@ public void downloadFile(String url, File to)
...
+        final int remainingUrl = urls.size() % threads;
+
+        for (int t = 0; t < threads; t++) {
+            final int thread = t;
+            threadss[t] = new Thread() {
+                @Override
+                public void run() {
+                    runThread(urls, threads, thread, UrlPerThread, remainingUrl);
+                }
+            };
```
1whatleytay, line 97:
```suggestion
            threadss[t] = new Thread(() -> { runThread(urls, threads, thread, UrlPerThread, remainingUrl) });
```

src/main/java/org/hydev/mcpm/client/Downloader.java
```diff
@@ -40,8 +83,45 @@ public void downloadFile(String url, File to)
      */
     public void downloadFiles(Map<String, File> urls, boolean progress, int threads)
     {
-        // TODO: Implement this
-        throw new UnsupportedOperationException("TODO");
+        Thread[] threadss = new Thread[threads];
```
1whatleytay, line 86: I'd try to pick names that are more identifying of what each variable does. In this function you have `t`, `thread`, `threads` and `threadss`... I might consider naming them `currentThreadId`, `threadCount` and `threadList`...

src/main/java/org/hydev/mcpm/client/Downloader.java
```diff
@@ -40,8 +83,45 @@ public void downloadFile(String url, File to)
...
+        final int UrlPerThread = urls.size() / threads;
```
kchprog, line 87: should this be in

src/main/java/org/hydev/mcpm/client/Downloader.java
```diff
@@ -17,16 +27,49 @@
...
+                    .desc(format("Downloaded", all.size()))
+                    .descLen(30);
+            all.add(b.appendBar(row));
+            while ((read = in.read(buffer, 0, 1024)) >= 0) {
+
+                bout.write(buffer, 0, read);
+                all.forEach(a -> a.increase(1024));
+            }
+            bout.close();
+            in.close();
```
1whatleytay, line 67: Consider using a try/using statement:
```java
try (FileOutputStream fileos = new FileOutputStream(to)) {
  ...
}
```

src/main/java/org/hydev/mcpm/client/Downloader.java
```diff
@@ -40,8 +83,45 @@ public void downloadFile(String url, File to)
      */
     public void downloadFiles(Map<String, File> urls, boolean progress, int threads)
```
kchprog, line 84: This needs a docstring, otherwise it looks fine.

src/main/java/org/hydev/mcpm/client/Downloader.java
```diff
@@ -17,16 +27,49 @@
...
+            BufferedInputStream in = new BufferedInputStream(http.getInputStream());
+            FileOutputStream fileos = new FileOutputStream(to);
+            BufferedOutputStream bout = new BufferedOutputStream(fileos, 1024);
+            byte[] buffer = new byte[1024];
+            int read = 0;
+
+            var b = new ProgressBar(ProgressBarTheme.ASCII_THEME);
+            var all = new ArrayList<ProgressRow>();
+            var row = new ProgressRow(fileSize)
+                    .unit("MB")
```
1whatleytay, line 57: Is `MB` correct here? You increment by `1024` every time you read `1024` bytes. Do you mean `"bytes"`?

src/main/java/org/hydev/mcpm/client/Downloader.java
```diff
@@ -40,8 +83,45 @@ public void downloadFile(String url, File to)
...
+        for (Map.Entry<String, File> url:urls.entrySet()) {
+            files.add(url.getKey());
+        }
+        // Store urls per thread
+        List<String> inFiles = new ArrayList<>();
+
+        for (int i=thread*UrlPerThread; i < (thread + 1) * UrlPerThread; i++) {
+            inFiles.add(files.get(i));
+        }
+        if (thread == threads-1 && remainingUrl > 0) {
```
1whatleytay, line 116: Just out of curiosity, what does `thread == threads-1` do here?

kchprog (commented):
Looks mostly fine. I'd suggest adding some more detailed docstrings and making sure the styling's correct.

src/main/java/org/hydev/mcpm/client/Downloader.java
```diff
@@ -40,8 +83,45 @@ public void downloadFile(String url, File to)
...
+            for (int j = files.size() - remainingUrl; j < files.size(); j++) {
+                inFiles.add(files.get(j));
+            }
+        }
+
+        for (String url:inFiles) {
+            downloadFile(url, urls.get(url));
```
1whatleytay, line 123: It looks like you have multiple ProgressBars for every thread.
 - Can you create one big `ProgressBar` for the whole operation and pass a `ProgressRow` down to `downloadFile`?
  - Are there any issues with multiple `ProgressBar`s working at the same time? Is `ProgressBar`s thread safe? 
  - You might need to lock `ProgressBar`s with a mutex or a `synchronized (bar) { ... }` block.

1whatleytay (approved):

src/main/java/org/hydev/mcpm/client/Downloader.java
MstrPikachu: there are a total of `threads` threads, so `thread == threads - 1` tests if this is the final iteration of the loop

1whatleytay: What loop? This method doesn't call itself does it...

MstrPikachu: I have a reading issue. But this method is called from a loop, which uses the same naming conventions, so I guess this means if this is the "final" thread

MstrPikachu: It's for if the # files isn't divisible by # threads, to add the remainder manually

src/main/java/org/hydev/mcpm/client/Downloader.java
```diff
@@ -40,8 +83,45 @@ public void downloadFile(String url, File to)
...
+        }
+
+        for (Thread t: threadss) {
+            t.start();
+        }
+    }
+    private void runThread(Map<String, File> urls, int threads, int thread, int UrlPerThread, int remainingUrl) {
+        // Store all urls into a list
+        List<String> files = new ArrayList<>();
+        for (Map.Entry<String, File> url:urls.entrySet()) {
```
MstrPikachu, line 107: As far as I can tell this isn't thread-safe, maybe add a synchronized block here.
But the bigger problem is that iteration over maps is not guaranteed to be in the same order every time.
The map is probably unchanging but you might run into a bug down the line.

MstrPikachu: What you could do instead of copying inside the thread is do the copying outside the thread, and then pass it in. This would avoid the concurrency issue I mentioned above

1whatleytay: What's not thread safe btw? The map isn't being mutated.

MstrPikachu: accessing is thread-safe? whoops

src/main/java/org/hydev/mcpm/client/Downloader.java
1whatleytay: Ahh gotcha. In an ideal world we could have a queue system so threads that are idle can pick up work instead of going to sleep. We wouldn't have to worry about splitting threads like this.

src/main/java/org/hydev/mcpm/client/Launcher.java
thudoan1706: I borrow the main class to test the file and deleted those when pushing from local XD it won't be the same previously

src/main/java/org/hydev/mcpm/client/Downloader.java
thudoan1706: noted

PR #28, opened 2022-11-14
Feature out of order progress bars

MstrPikachu:
Progress bars can now finish out of order
still scuffed if there are more bars than height of terminal


1whatleytay (dismissed):

hykilpikonna (dismissed):

MstrPikachu:
wait it doesn't work yet

src/main/java/org/hydev/mcpm/client/interaction/ProgressBarBoundary.java
```diff
@@ -0,0 +1,50 @@
...
+     *
+     * @param bar Progress bar
+     */
+    void finishBar(ProgressRowBoundary bar);
+
+    /**
+     * Finalize and close the progress bar (print the final line)
+     */
+    @Override
+    void close();
```
1whatleytay, line 32: Why is this overridden and do we actually override this method with anything useful? Seems empty to me.

1whatleytay (approved):

src/main/java/org/hydev/mcpm/client/interaction/ProgressBarBoundary.java
MstrPikachu: My bad I for some reason added this here even though it gets inherited from AutoCloseable

hykilpikonna: Yea for single-line progress bars, since they are using `\r` to reset the cursor to col 1 on each update, it needs to print a new line when the progress bar is done (i.e. closed). But I figured that multi-line progress bars don't need this... maybe we should remove it from the boundary?

hykilpikonna (approved):

PR #53, opened 2022-11-20
Implement installer

thudoan1706:


src/main/java/org/hydev/mcpm/client/database/inputs/SearchPackagesInput.java
```diff
@@ -10,13 +10,6 @@
...
-    /**
-     * The possible types of searching.
-     */
-    public enum Type {
-        BY_NAME,
-        BY_COMMAND,
-        BY_KEYWORD
-    }
+public record SearchPackagesInput(SearchPackagesType type, String searchStr, boolean noCache) {
+
```
MstrPikachu: uh what happened here

MstrPikachu: it got moved into another class?

1whatleytay: Yep.

src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
```diff
@@ -0,0 +1,99 @@
...
+    * 4. Download it (Downloader)
+    * 5. Add plugin tracker to check install and write
+    * 6. Install the list of dependency
+    *
+    * */
+
+        List<PluginYml> pluginInstalled = installInput.pluginTracker().listInstalled();
+        for (PluginYml pluginYml:pluginInstalled) {
+            if (pluginYml != null && pluginYml.name() != null && pluginYml.name().equals(pluginVersion.meta().name())) {
+                throw new RuntimeException();
```
MstrPikachu: This will error if there are dependencies that are installed locally already.
In the context of the install use case, you'd want to skip over dependencies that are already installed.
In this case, return would be more appropriate than an exception.

MstrPikachu: I read it wrong, it'll error if you're trying to install an already installed plugin, but in that case the exception should tell you that instead of having a blank message

MstrPikachu: I take it back
Every dependency is being recursed on, instead of only those that are not installed, meaning that already installed dependencies will throw an exception. This should be a return instead (recursive base case)

1whatleytay: Oh yep, this is a bug. Although this can be caught. The best solution here would be to make an `InstallResult` object that has one of those state variables. This will kinda force us to deal with these hard questions (when to throw when not to).

MstrPikachu: I don't see why we should catch only to return an error state when we can directly return the error state.
But also remember that fundamentally, DFS returns once for every visit (you can't just return the recursive call, as this would not continue the other required recursive calls).
So do you want to make a state for "no dependencies already installed" and "some dependencies already installed"?
I think it makes more sense to return a list of all dependencies that are already installed, if we care about those. Otherwise I would keep the method void and just put a return here.

src/main/java/org/hydev/mcpm/client/installer/SpigotPluginDownloader.java
```diff
@@ -0,0 +1,32 @@
...
+import java.io.File;
+
+public class SpigotPluginDownloader implements PluginDownloader {
+
+    private final Downloader downloader;
+    private final String baseURL;
+
+    public SpigotPluginDownloader(Downloader downloader) {
+        this.downloader = downloader;
+        this.baseURL = "https://mcpm.hydev.org";
```
hykilpikonna:
```suggestion
        this(downloader, "https://mcpm.hydev.org")
```

hykilpikonna (approved):

src/main/java/org/hydev/mcpm/client/adaptor/DownloaderProgressBar.java
```diff
@@ -0,0 +1,26 @@
...
+
+    }
+
+    @Override
+    public void incrementBarProgress(int id, long inc) {
+        List<ProgressRowBoundary> bars = getBars();
+        bars.get(id).increase(inc);
+    }
+
+}
```
hykilpikonna: why are separate progress bars needed?

thudoan1706: Oh. I actually didn't use it. I was planning to change some of the speed, but I don't need it anymore. I will delete this.

hykilpikonna: okay great!

hykilpikonna (approved):

1whatleytay (approved):
When conflicts are fixed: Merge it!

hykilpikonna:
> When conflicts are fixed: Merge it!

There are so many conflicts wth

PR #56, opened 2022-11-20
Rework installer

thudoan1706:


src/main/java/org/hydev/mcpm/client/installer/InstallController.java
```diff
@@ -0,0 +1,75 @@
...
+import org.hydev.mcpm.client.database.fetcher.LocalDatabaseFetcher;
+import org.hydev.mcpm.client.database.inputs.SearchPackagesType;
+import org.hydev.mcpm.client.database.results.SearchPackagesResult;
+import org.hydev.mcpm.client.installer.input.InstallInput;
+import org.hydev.mcpm.client.models.PluginModel;
+
+import java.io.File;
+import java.net.URI;
+
+public class InstallController implements InstallBoundary {
```
jerryzhu509: I don't believe that the use case controller should implement InstallBoundary. At a high level, the controller should be reading the input from the user and wrapping it into an InstallInput object, which you pass forward to the InstallIBoundary. The InstallBoundary is implemented by InstallInteractor, which controls the "dance of entities" as Professor Gries says in class, spitting out a result.

TLDR: I think the controller shouldn't implement installboundary; installinteractor should. Refactor by moving this overridden method into the interactor.

src/main/java/org/hydev/mcpm/client/installer/InstallController.java
```diff
@@ -0,0 +1,75 @@
...
+        SearchPackagesResult searchPackagesResult = installInteractor.searchPackageInput(installInput);
+        if (searchPackagesResult.plugins().isEmpty()) {
+            throw new RuntimeException();
+        }
+
+        PluginModel latestPluginModel = searchPackagesResult.plugins().get(0);
+
+        long id = latestPluginModel.id();
+
+        for (PluginModel plugin: searchPackagesResult.plugins()) {
```
jerryzhu509: Case in point: Controller, being in the adapters layer, should not have direct access to the entities layer. It violates clean architecture for your controller to deal directly with entities. These are actions that should go into the use case interaction, not the controller.

src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
```diff
@@ -12,88 +12,138 @@
...
-    public void installPlugin(InstallInput installInput) {
-        SearchPackagesInput searchPackagesInput = new SearchPackagesInput(installInput.type(), installInput.name(), false);
-        SearchPackagesResult searchPackageResult = installInput.searchPackage().search(searchPackagesInput);
-        if (searchPackageResult.state() != SearchPackagesResult.State.SUCCESS) {
-            throw new IllegalArgumentException("Illegal");
-        }
+/*
+* An interactor where it mainly communicates with database
+* */
+public class InstallInteractor{
```
jerryzhu509: Per my previous point, the interactor should be the one implementing the boundary and engaging with the entities.

src/main/java/org/hydev/mcpm/client/installer/InstallController.java
```diff
@@ -0,0 +1,86 @@
...
+ * Install Controller which handles the install input and request the database to
+ * serve installing use case
+ */
+public class InstallController implements InstallBoundary {
+    private final InstallInteractor installInteractor;
+    private final SpigotPluginDownloader spigotPluginDownloader;
+    private final Downloader downloader;
+
+
+    public InstallController() {
```
jerryzhu509: Need javadoc.

src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
```diff
@@ -12,88 +12,154 @@
...
-        InstallInput installInput = new InstallInput("JedCore", SearchPackagesType.BY_NAME, "",
-                true, databaseInteractor, spigotPluginDownloader, localPluginTracker);
-        installInteractor.installPlugin(installInput);
-    }
-}
+
+
+
+
+//    @Override
```
jerryzhu509: Remove the commented-out code if not needed.

src/main/java/org/hydev/mcpm/client/installer/PluginDownloader.java
```diff
@@ -1,5 +1,11 @@
 package org.hydev.mcpm.client.installer;
 
+/**
+ * Interface for plugin downloader.
+ * @since 2022-11-18
+ */
 public interface PluginDownloader {
-    public void download (long pluginId, long pluginVersion, String destination);
+    public void download(long pluginId,
```
jerryzhu509: Comments needed

src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
```diff
@@ -12,88 +12,154 @@
...
-        for (PluginModel plugin: searchPackageResult.plugins()) {
-            if (plugin.id() > id) {
-                id = plugin.id();
-                latestPluginModel = plugin;
+    /**
+     * Search if the plugin exists, throw exception if it does not exist and return the information of
+     * the pluigin needs to be installed.
+     * @param installInput: the plugin input pending for installation
+     */
+    public SearchPackagesResult searchPackageInput(InstallInput installInput) {
```
jerryzhu509: Consider renaming the method name to an action instead of a noun (e.g. getSearchResults) in compliance with java naming conventions.

src/main/java/org/hydev/mcpm/client/installer/InstallController.java
```diff
@@ -0,0 +1,86 @@
...
+        long id = latestPluginModel.id();
+        for (PluginModel plugin: searchPackagesResult.plugins()) {
+            if (plugin.id() > id) {
+                id = plugin.id();
+                latestPluginModel = plugin;
+            }
+        }
+        var pluginLatestVersion = latestPluginModel.getLatestPluginVersion();
+        if (pluginLatestVersion.isEmpty()) {
+            throw new RuntimeException();
```
jerryzhu509: Might want to throw something more specific for debugging ease. Make a comment on the side for a remainder if you don;t want to deal with it for now.

src/main/java/org/hydev/mcpm/client/installer/InstallController.java
```diff
@@ -0,0 +1,86 @@
...
+    * Install the plugin
+    * @param installInput: the plugin submitted by the user for installing
+    * */
+    @Override
+    public void installPlugin(InstallInput installInput) {
+        SearchPackagesResult searchPackagesResult = installInteractor.searchPackageInput(installInput);
+
+        // Check if the searched plugin for installing is found
+        if (searchPackagesResult.plugins().isEmpty()) {
+            throw new RuntimeException();
```
jerryzhu509: Might want to throw something more specific for debugging ease. Make a comment on the side for a remainder if you don;t want to deal with it for now.

jerryzhu509 (changes requested):

src/main/java/org/hydev/mcpm/client/installer/InstallController.java
thudoan1706: Yes! Thank you for pointing that out. I was a bit confused about this previously as well, and it doesn't seem right that the controller would be at the same level as the use case interactor. 

I just fixed it, you can help me take a look if there's something off.

src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
thudoan1706: solved

src/main/java/org/hydev/mcpm/client/installer/PluginDownloader.java
thudoan1706: solved.

src/main/java/org/hydev/mcpm/client/installer/DatabaseManager.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * Search if the plugin exists, throw exception if it does not exist and return the information of
+     * the pluigin needs to be installed.
+     * @param installInput: the plugin input pending for installation
+     */
+    public SearchPackagesResult getSearchResult(InstallInput installInput) {
+        SearchPackagesInput searchPackagesInput = new SearchPackagesInput(installInput.type(), installInput.name(), false);
+        SearchPackagesResult searchPackageResult = databaseInteractor.search(searchPackagesInput);
```
jerryzhu509: Note that this will need to change later since DatabaseInteractor should be separated into distinct use cases for clean architecture. In particular, my new PR makes a SearchInteractor class for searching instead of using db interactor.

src/main/java/org/hydev/mcpm/client/installer/DatabaseManager.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.io.File;
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.List;
+
+/*
+* An interactor that communicates with database and returns the plugin source for installation
+* */
+public class DatabaseManager {
```
jerryzhu509: A bit confused.. If this is also an interactor, then where was the controller from before to convert inputs? While classes on the same level can depend on each other, we do want to minimize the dependencies.

src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
```diff
@@ -5,95 +5,81 @@
...
+/**
+ * Install Interactor which serves the use case of installing plugin
+ */
 public class InstallInteractor implements InstallBoundary {
+    private final DatabaseManager databaseManager;
+    private final SpigotPluginDownloader spigotPluginDownloader;
+    private final Downloader downloader;
+
+
+    public InstallInteractor() {
```
jerryzhu509: comments needed

src/main/java/org/hydev/mcpm/client/installer/DatabaseManager.java
```diff
@@ -0,0 +1,82 @@
...
+                throw new RuntimeException();
+            }
+        }
+    }
+
+    /**
+    * Add the installed plugin to the csv file
+    * @param pluginVersion: The version of the installed plugin
+    * */
+    public void addManualInstalled(PluginVersion pluginVersion) {
```
jerryzhu509: This does indeed seem like a use case method. If I understand correctly, this should be all for the installing use case. Why the need for two intearctors? That is, why not put this in installinteractor? You'll also need to create another boundary for this is an interactor.

src/main/java/org/hydev/mcpm/client/installer/InstallController.java
jerryzhu509: Great work on the implementation! I just have a couple of followup questions most likely just due to my not fully understanding your work. Also make sure to check up on the errors gradle checkstyle is pointing out to you in Actions

src/main/java/org/hydev/mcpm/client/installer/DatabaseManager.java
thudoan1706: You have a good point. I was trying to let the installer interactor invokes the install method to serve its single-use case. If we add this add manual install, then it has to couple with more classes in the database and carry more responsibility by working on with database. So I just create another hook for the interactor (database manager) to check and prepare the input for the interactor. Another thing I found is that what if uninstall and update have to access to the database to carry out the same check like install. Then, I think a good choice is to separate them, and yes I would need an interface for the database manager if I do so. Feel free to correct my thought :D

1whatleytay (approved):
Merge it (after comments are addressed)!

jerryzhu509 (approved):

src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
```diff
@@ -57,38 +55,31 @@ public void installPlugin(InstallInput input) throws InstallException
         }
 
         var pluginVersion = latestPluginModel.getLatestPluginVersion()
-            .orElseThrow(() -> new InstallException(Type.NO_VERSION_AVAILABLE));
+                .orElseThrow(() -> new InstallException(Type.NO_VERSION_AVAILABLE));
+        databaseManager.checkPluginInstalled(pluginVersion);
```
MstrPikachu, line 71: I see you took out the code that used to be here and put it into a method

But there's still a problem that if a dependency (including the initial plugin) is already installed, the rest of the installation will abort because of the thrown exception. 
In terms of the expected behaviour for the use case, we would want it to skip over these plugins, not to stop the installation.
Taylor suggested that we can return the state of the entire installation. The states could be like this: success and fail. Success is if EVERY installation either succeeds or is skipped because it is already installed (NOT a fail). A fail is if there is an error somehow, for AT LEAST one of the plugins.

You could do something like this (here I will use true and false for success and fail, but you should wrap it with a Record):
```suggestion
        if (searchResult.plugins().isEmpty()) {
             // throw new InstallException(Type.NOT_FOUND);
             return false; // we failed so return false
        }
        
...

       var name = pluginVersion.meta().name();
         List<PluginYml> pluginInstalled = input.pluginTracker().listInstalled();
         for (PluginYml pluginYml : pluginInstalled) {
             if (pluginYml != null && pluginYml.name() != null && pluginYml.name().equals(name)) {
                 return true; // we skip the installed plugin so return true because we did not fail
             }
         }

...

        // 4. Installing the depency of that plugin
         boolean ret = true;
         if (pluginVersion.meta().depend()!= null) {
             for (String dependency:pluginVersion.meta().depend()) {
                 InstallInput dependencyInput = new InstallInput(dependency,
                         SearchPackagesType.BY_NAME,
                         installInput.filePath(),
                         installInput.load());
                 ret &= installPlugin(dependencyInput); // we return true iff ALL installs succeed (do not fail), so we use and
 //                input.pluginTracker().removeManuallyInstalled(name);
             }
         }
         return ret;
```

MstrPikachu (dismissed):

src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
MstrPikachu: Obviously you can't do &= with a record. But what you can do is:
```suggestion
public record InstallResult
{
    State state, // this corresponds to whether failedPlugins is empty or not
    Collection<PluginModel> failedPlugins;
    public enum State {
        SUCCESS,
        FAILED
    }
    public void merge(InstallResult other) {
        if (this.state == State.SUCCESS && other.state == State.SUCCESS) {
            return;
        }
        state.State = State.FAILED;
        failedPlugins.addAll(other.failedPlugins);
}
```
and then in the install method:
```suggestion
        // 4. Installing the depency of that plugin
         InstallResult ret = new InstallResult(InstallResult.State.SUCCESS, List.of());
         if (pluginVersion.meta().depend()!= null) {
             for (String dependency:pluginVersion.meta().depend()) {
                 InstallInput dependencyInput = new InstallInput(dependency,
                         SearchPackagesType.BY_NAME,
                         installInput.filePath(),
                         installInput.load());
                 ret.merge(installPlugin(dependencyInput)); // we return true iff ALL installs succeed (do not fail), so we use and
 //                input.pluginTracker().removeManuallyInstalled(name);
             }
         }
         return ret;
```

src/main/java/org/hydev/mcpm/client/installer/DatabaseManager.java
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * Search if the plugin exists, throw exception if it does not exist and return the information of
+     * the pluigin needs to be installed.
+     * @param input: the plugin input pending for installation
+     */
+    public SearchPackagesResult getSearchResult(InstallInput input) throws InstallException {
+        SearchPackagesInput searchPackagesInput = new SearchPackagesInput(input.type(), input.name(), false);
+        SearchPackagesResult searchPackageResult = searchInteractor.search(searchPackagesInput);
+        if (searchPackageResult.state() != SearchPackagesResult.State.SUCCESS) {
+            throw new InstallException(
```
MstrPikachu: I mentioned this already in the install method, but I would probably want to have all the plugins installed even if some can't be searched for some reason. You can do something similar with my suggestion for checkPluginInstalled

But if you believe that you should abort early if some packages can't be searched for, this is fine.

src/main/java/org/hydev/mcpm/client/installer/InstallController.java
1whatleytay: Is this true? We can reach down layers we just can't reach up given DI.

hykilpikonna (approved):

hykilpikonna:
Great work everyone!

PR #57, opened 2022-11-20
uninstall_usecase

aanushkasharma:
Uninstall method in uninstaller class currently should only unload and uninstall one plugin as of now; removing plugin dependencies still needs to be implemented.


src/main/java/org/hydev/mcpm/client/uninstallUseCase/Uninstaller.java
```diff
@@ -0,0 +1,36 @@
...
+        PluginLoader pil = new PluginLoader();
+        pil.unloadPlugin(name);
+
+        // Find plugin file
+        // could be useful for removing dependencies (?)
+        var plugins_folder = new File("plugins");
+        var plugins = Arrays.stream((plugins_folder.listFiles()))
+                .filter(plugin -> plugin.getName().endsWith(name+".jar"));
+
+        File file = new File(plugins.toArray()[0].toString());
```
1whatleytay, line 22:
```suggestion
        Optional<File> fileThatHasPluginThatMightNotExist = plugin.findFirst();
        
       if (fileThatHasPluginThatMightNotExist.isEmpty()) {
           // ... print error message that there is no plugin installed like this? ...
           return;
       }
        
        File file = fileThatHasPluginThatMightNotExist.get();
```

src/main/java/org/hydev/mcpm/client/uninstallUseCase/Uninstaller.java
```diff
@@ -0,0 +1,36 @@
...
+import org.hydev.mcpm.client.injector.PluginLoader;
+import org.hydev.mcpm.client.injector.PluginNotFoundException;
+
+
+public class Uninstaller implements UninstallBoundary{
+    private boolean remove_depen;
+
+
+    public void uninstall(String name) throws PluginNotFoundException {
+        PluginLoader pil = new PluginLoader();
```
1whatleytay, line 13: This is concrete, you should take an `UnloadBoundary` as part of an `UninstallInput` object instead of creating a plugin loader like this.

Generally we want the uninstall method to work with *any* way to unload plugins, so if we want to test it or if we decide to switch how we want to unload stuff it becomes super easy.

src/main/java/org/hydev/mcpm/client/uninstallUseCase/Uninstaller.java
```diff
@@ -0,0 +1,36 @@
...
+        File file = new File(plugins.toArray()[0].toString());
+
+        if (file.delete()) {
+            System.out.println("File deleted successfully");
+        } else {
+            System.out.println("Failed to delete the file");
+        }
+
+
+//        if (remove_depen){
```
1whatleytay, line 31: Just for reference, `PluginYML` has a `depend` array containing the names of all the plugins it depends on. Before removing them outright(!) though I'd check PluginTracker to see if any of the dependencies are in `listManuallyInstalled` (because if the user installed them we definitely don't want to get rid of them as like a tangent).

src/main/java/org/hydev/mcpm/client/uninstallUseCase/Uninstaller.java
```diff
@@ -0,0 +1,36 @@
...
+        var plugins = Arrays.stream((plugins_folder.listFiles()))
+                .filter(plugin -> plugin.getName().endsWith(name+".jar"));
+
+        File file = new File(plugins.toArray()[0].toString());
+
+        if (file.delete()) {
+            System.out.println("File deleted successfully");
+        } else {
+            System.out.println("Failed to delete the file");
+        }
```
1whatleytay, line 28: You're going to have to go through PluginTracker to mark the plugin as an input (take this as an input). You can probably use `removeManuallyInstalled`.

1whatleytay (commented):
Sorry, I know this is work in progress but I left some comments here if you're looking for what to do next.

If you know what you're doing just ignore them for now.

src/main/java/org/hydev/mcpm/client/uninstallUseCase/UninstallInput.java
```diff
@@ -0,0 +1,9 @@
+package org.hydev.mcpm.client.uninstallUseCase;
+
```
kchprog, line 2: Seeing as this is unfinished, an implementation of the input as a record should work.

src/main/java/org/hydev/mcpm/client/uninstallUseCase/Uninstaller.java
```diff
@@ -0,0 +1,36 @@
...
+    private boolean remove_depen;
+
+
+    public void uninstall(String name) throws PluginNotFoundException {
+        PluginLoader pil = new PluginLoader();
+        pil.unloadPlugin(name);
+
+        // Find plugin file
+        // could be useful for removing dependencies (?)
+        var plugins_folder = new File("plugins");
```
kchprog, line 18: This information should be static. There is a default folder containing the plugins at the relative location "plugins", and there might already be a method that does this work.

src/main/java/org/hydev/mcpm/client/uninstallUseCase/Uninstaller.java
kchprog, line 22: What if the plugin file is empty? Make sure edge cases and exceptions are handled whenever possible.

kchprog (changes requested):
Please consider implementing some of these suggestions, and pushing the other contributions to the branch for reviewing.

src/main/java/org/hydev/mcpm/client/uninstallUseCase/Uninstaller.java
thudoan1706: I am thinking of making a class to handle the file system (maybe the local tracker can handle this) for installation, uninstall and update.

1. create a folder to store plugins
   - If a folder doesn't exist: 
      + Create one for installing (install)
      +  Send an error message for an update (since the plugin does not install yet) and uninstall (the plugin does not exist)
    - If it exists, do nothing

2.Check (and store) the file path of the (older version) plugins in the directory for uninstall and delete
    For updates, use uninstall and bypass the file path to install input => install. The parenthesis note is for the update.

kchprog (changes requested):
Please consider comments, and please push new files as soon as possible. Thanks.

src/main/java/org/hydev/mcpm/client/uninstallUseCase/Uninstaller.java
MstrPikachu: I think you also want to return the result of the uninstall, for example success or some kind of failure.
You can do this by returning an UninstallResults record.

src/main/java/org/hydev/mcpm/client/uninstallUseCase/Uninstaller.java
MstrPikachu: You also want to make it recursive, in case the dependencies have dependencies
Your base case would be if the plugin is not installed, then you don't have to do recursion.
Although this wouldn't work if you manually uninstalled something and wanted to actually uninstall all its plugins.
In that case you probably want a visited array in case of circular dependencies, then the base case would be to return if you've already visited the plugin.

MstrPikachu: Iterative works as well, the benefit to iterative is that you may not have to make a helper recursive function, but also could be harder to implement. lmk if I can help

src/main/java/org/hydev/mcpm/client/uninstallUseCase/UninstallInput.java
thudoan1706: We can use the getSearchReult method in the database manager in the installer folder to help you check if the plugin exists or not.

hykilpikonna:
Closed because new pr opened #82

PR #59, opened 2022-11-20
[+] Color logger

hykilpikonna:
* Add the basic framework for colored logging in both Minecraft and Console environments
* Add an `uncheckedMapOf()` method to create infinitely-long hash maps whereas the checked `Map.of` has limited number of arguments
* Make `ArgParser` work better with custom loggers

image

### DIFF
--- a/src/main/java/org/hydev/mcpm/client/arguments/ArgsParser.java
+++ b/src/main/java/org/hydev/mcpm/client/arguments/ArgsParser.java
@@ -1,13 +1,14 @@
 package org.hydev.mcpm.client.arguments;
 
 import net.sourceforge.argparse4j.ArgumentParsers;
-import net.sourceforge.argparse4j.inf.ArgumentParser;
-import net.sourceforge.argparse4j.inf.ArgumentParserException;
+import net.sourceforge.argparse4j.impl.Arguments;
+import net.sourceforge.argparse4j.inf.*;
 import org.hydev.mcpm.client.arguments.parsers.CommandParser;
 
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.util.List;
+import java.util.Map;
 import java.util.function.Consumer;
 
 /**
@@ -35,12 +36,31 @@ public class ArgsParser
         var parsers = parser.addSubparsers();
 
         for (var parser : allParsers) {
-            var subparser = parser.configure(parsers);
+            var subparser = parsers.addParser(parser.name(), false);
+            parser.configure(subparser);
 
-            if (subparser != null) {
-                subparser
-                    .setDefault("handler", parser);
+            subparser.setDefault("handler", parser);
+
+            class PrintHelpAction implements ArgumentAction
+            {
+                @Override
+                public void run(ArgumentParser parser, Argument arg, Map<String, Object> attrs,
+                                String flag, Object value) {
+                    log.accept(subparser.formatHelp());
+                }
+
+                @Override
+                public void onAttach(Argument arg) {
+                    /* nothing */
+                }
+
+                @Override
+                public boolean consumeArgument() {
+                    return false;
+                }
             }
+
+            subparser.addArgument("-h", "--help").action(new PrintHelpAction());
         }
     }
 
--- a/src/main/java/org/hydev/mcpm/client/arguments/ArgsParser.java
+++ b/src/main/java/org/hydev/mcpm/client/arguments/ArgsParser.java
@@ -30,10 +30,7 @@ public class ArgsParser
     public ArgsParser(List<CommandParser> allParsers, Consumer<String> log) {
         this.log = log;
 
-        parser = ArgumentParsers
-            .newFor("loader")
-            .build()
-            .defaultHelp(true);
+        parser = ArgumentParsers.newFor("mcpm").build().defaultHelp(true);
 
         var parsers = parser.addSubparsers();
 
--- a/src/main/java/org/hydev/mcpm/client/arguments/ArgsParser.java
+++ b/src/main/java/org/hydev/mcpm/client/arguments/ArgsParser.java
@@ -57,8 +57,7 @@ public class ArgsParser
      *                                 For default handling, pass this to ArgsParser#fail.
      */
     public void parse(String[] arguments, Consumer<String> log) throws ArgumentParserException {
-        try
-        {
+        try {
             var namespace = parser.parseArgs(arguments);
 
             Object handleObject = namespace.get("handler");
@@ -69,8 +68,7 @@ public class ArgsParser
 
             handler.run(namespace, log);
         }
-        catch (HelpException e)
-        {
+        catch (HelpException e) {
             log.accept(e.help());
         }
     }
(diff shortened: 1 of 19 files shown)


src/main/java/org/hydev/mcpm/client/arguments/ArgsParser.java
```diff
@@ -30,20 +30,16 @@
     public ArgsParser(List<CommandParser> allParsers, Consumer<String> log) {
         this.log = log;
 
-        parser = ArgumentParsers
-            .newFor("loader")
-            .build()
-            .defaultHelp(true);
+        parser = ArgumentParsers.newFor("mcpm").build().defaultHelp(true);
```
1whatleytay, line 33: I like the newline formatting. I was gonna comment on how you squished everything to one line on the lower builders but I see you've changed existing stuff too 😦 Not a big deal and definitely nothing like a merge blocker, but I'm sad.

src/main/java/org/hydev/mcpm/client/arguments/ArgsParser.java
```diff
@@ -61,15 +57,20 @@ public void parse(String[] arguments) throws ArgumentParserException {
...
-            throw new ArgumentParserException("Unrecognized command.", parser);
-        }
+            if (!(handleObject instanceof CommandParser handler)) {
+                throw new ArgumentParserException("Unrecognized command.", parser);
+            }
 
-        handler.run(namespace, log);
+            handler.run(namespace, log);
+        }
+        catch (HelpException e) {
```
1whatleytay, line 71: This can't be done inside the action because... it will continue parsing arguments? Oh I see, when we set consume arguments to false, we had `--help` get consumed as the text argument. 😳 Gotcha, nice solution!

src/main/java/org/hydev/mcpm/utils/ColorLogger.java
```diff
@@ -0,0 +1,98 @@
...
+        // Add &r (color reset) after each line
+        //in = String.join("\n", Arrays.stream(in.split("\n")).map(line -> line + "&r").toList());
+
+        // Add &r to the end
+        in += "&r";
+
+        // Replace colors
+        for (var entry : ansiReplacements.entrySet())
+        {
+            in = in.replace(entry.getKey(), entry.getValue());
```
1whatleytay, line 76: Does Minecraft have escape strings? What if I want to log plaintext `this&a pizza`?

A template library is usually best to format these things. Obviously with our limited time we can merge this but if this was production this would find a way to bite us pretty quickly I think.

src/main/java/org/hydev/mcpm/utils/ColorLogger.java
```diff
@@ -0,0 +1,98 @@
...
+        "&b", "\033[1;36m",
+        "&c", "\033[1;31m",
+        "&d", "\033[1;35m",
+        "&e", "\033[1;33m",
+        "&f", "\033[1;37m",
+        "&r", "\033[0m",
+        "&l", "\033[1m",
+        "&o", "\033[3m",
+        "&n", "\033[4m",
+        "&-", "\n"
```
1whatleytay, line 41: Also do we have a logging library that can handle this for us? Logging libraries should be responsible to handle escape sequences in a proper way. Some terminals (notably Xcode) don't support these escapes and the result is a bunch of garbled text.

1whatleytay (approved):

src/main/java/org/hydev/mcpm/utils/ColorLogger.java
hykilpikonna: Check out https://github.com/HyDevelop/HyLogger (I made it lol)

Also [HyFetch can try to detect](https://github.com/hykilpikonna/hyfetch/blob/b86f43e9abc6e7b618eedb06d889fe404d55b886/hyfetch/termenv.py#L18) if a terminal supports Xterm256 or Xterm truecolor but not if it doesn't support ANSI.

But tbh there's really no way of correctly identifying if a terminal supports ANSI that works 100% of the time. There are so many different terminals out there and not all of them follow the same standard.

hykilpikonna: I'll just assume that whoever is using the tool on the command line is using it with a terminal that supports colors.

src/main/java/org/hydev/mcpm/utils/ColorLogger.java
hykilpikonna: Yea in minecraft you can use `&&a` to escape `&a` but I don't have time to implement that now 😭

src/main/java/org/hydev/mcpm/client/arguments/ArgsParser.java
hykilpikonna: oops sorry

src/main/java/org/hydev/mcpm/utils/ColorLogger.java
1whatleytay: > There are so many different terminals out there and not all of them follow the same standard.

This is why we move the responsibility to a library maintained by other people. If a terminal is not supported, then it's their fault to support it. As developers we can't possibly think of all the different environments our app will run in, but the library developers can do that thinking for us.

Again not a blocker. It's the last day before feedback lol

src/main/java/org/hydev/mcpm/utils/ColorLogger.java
1whatleytay: Spooky.

src/main/java/org/hydev/mcpm/utils/ColorLogger.java
hykilpikonna: Quick fix? 🤔 

```java
var ampEscape = "[gFk38N-EsCaPeD AmPeRcEnT-f47Svw]";
in = in.replace("&&", ampEscape);

// Replace colors

in = in.replace(ampEscape, "&");
```

src/main/java/org/hydev/mcpm/utils/ColorLogger.java
hykilpikonna: Quick fix? 🤔 

```java
var ampEscape = "[gFk38N EsCaPeD AmPeRsAnD f47Svw]";
in = in.replace("&&", ampEscape);

// Replace colors

in = in.replace(ampEscape, "&");
```

1whatleytay: Hahaha, you don't need to... lol

1whatleytay: Why do I keep commenting right after things get merged...

src/main/java/org/hydev/mcpm/utils/ColorLogger.java
MstrPikachu: > If a terminal is not supported, then it's their fault to support it. 

when you use intuition to skip steps when showing work and lose marks:
"it's not my fault you need more steps"

src/main/java/org/hydev/mcpm/utils/Sugar.java
```diff
@@ -54,4 +56,29 @@ public static <T> T[] subTo(T[] arr, int end)
...
+        {
+            var k = values[i];
+            var v = values[i + 1];
+
+            assert k != null;
+            if (v != null) map.put(k, v);
+        }
+
+        //noinspection unchecked
+        return (Map<K, V>) map;
```
MstrPikachu, line 82: Do we have to suppress any warnings here?

src/main/java/org/hydev/mcpm/utils/ColorLogger.java
1whatleytay: Man, this is a library. It's not like omitting steps 3 through 5, it's like citing a proven theorem.

src/main/java/org/hydev/mcpm/utils/Sugar.java
hykilpikonna: it's an unchecked cast. Due to type erasure, we cannot check if the keys or values are objects of type K or V, we can only put it in the precondition of the function and assume it's true. But it's still way better than Java's `Map.of` that only supports up to 10 pairs because their key and values are hardcoded

MstrPikachu (approved):
Good if there are no warnings

PR #60, opened 2022-11-21
[+] Better commands

hykilpikonna:
Better command framework

* Fix many issues with help message and loggers
* A more intuitive custom help message
* List command
* Util method for printing tables
* Single-level tab complete (naive)

image
image
image


src/main/java/org/hydev/mcpm/SpigotEntry.java
```diff
@@ -66,4 +72,18 @@ public boolean onCommand(@NotNull CommandSender sender, @NotNull Command command
...
+                                      @NotNull String[] args)
+    {
+        if (!command.getName().equalsIgnoreCase("mcpm")) return null;
+
+        if (args.length == 1)
+        {
+            return new ArrayList<>(parser.getRawSubparsers().stream().map(CommandParser::name).toList());
+        }
+
+        return null;
```
MstrPikachu, line 87: Should we be returning null instead of empty list? Surely it doesn't matter

hykilpikonna: Spigot supports returning null so it doesn't really matter

src/main/java/org/hydev/mcpm/client/arguments/ArgsParser.java
```diff
@@ -61,15 +70,23 @@ public void parse(String[] arguments) throws ArgumentParserException {
      *                                 For default handling, pass this to ArgsParser#fail.
      */
     public void parse(String[] arguments, Consumer<String> log) throws ArgumentParserException {
-        var namespace = parser.parseArgs(arguments);
+        // If no args are present, add help
+        if (arguments.length == 0) arguments = new String[]{"help"};
```
1whatleytay: Just have this print out help instead of going through ArgsParser.

1whatleytay (approved):

aanushkasharma (approved):

PR #67, opened 2022-11-21
Test progress speed classes

MstrPikachu:


hykilpikonna (approved):

src/main/java/org/hydev/mcpm/utils/arrays/SlidingWindow.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    public long sum() {
+        return sum;
+    }
+
+    @Override
+    public long sum(long index) {
+        restore(index);
+        return sum;
+    }
```
1whatleytay, line 41: Not that I dislike this but overloading to me implies that you have similar behaviour (especially when implementing methods in an interface). I would maybe suggest going with the traditional `getSum`/`setSum` combo.

src/test/java/org/hydev/mcpm/utils/arrays/SlidingWindowTest.java
```diff
@@ -0,0 +1,47 @@
...
+            if (arr.get(i).getKey() < index - window)
+                break;
+            ret += arr.get(i).getValue();
+        }
+        return ret;
+    }
+
+    @Test
+    void testAll() {
+        Random r = new Random();
```
1whatleytay, line 30: We can't have random numbers in a Unit Test like this. This makes the unit tests not deterministic (one call they might succeed another they might now). What you can do is pick a seed and put those numbers in List in a `@BeforeAll` method.

1whatleytay (approved):

src/test/java/org/hydev/mcpm/utils/arrays/SlidingWindowTest.java
MstrPikachu: well it should always succeed though

src/main/java/org/hydev/mcpm/utils/arrays/SlidingWindow.java
MstrPikachu: I agree but to me the behaviour does seem similar, `sum` does the same as `sum(index)` with the most recent index. Also `setSum` shouldn't be allowed because it's private.

Maybe you're suggesting that I violated the interface segregation principle which is probably true if I look at it again, but then it would be an issue about the design and not the tests anymore. I can fix/make an issue for that afterwards

src/test/java/org/hydev/mcpm/utils/arrays/SlidingWindowTest.java
MstrPikachu: I do get the idea that you might want the internal behaviour to always be identical, I think this makes the test more "thorough" although harder to debug
If you want I could have it store the seed somewhere so we can see or set it to a fixed seed

PR #68, opened 2022-11-22
[+] zstd compression

hykilpikonna:
* Add zstd decompression for database (reduces the db size from 21 MB to 3.6 MB
* Conditionally import native zstd by reflection (detect whether zstd is supported on the platform)
* Add fallback pure-java implementation for platforms that doesn't support native zstd
* Automatic systemd script to generate database

image

image


src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
```diff
@@ -153,7 +155,7 @@ private String readDatabaseFromContent(HttpEntity entity, DatabaseFetcherListene
         listener.finish();
 
         // Decompress ZSTD
-        if (this.enableCompression)
```
1whatleytay, line 156: Oh god... I think this is an IntelliJ warning (checking a value that's always true). Did you silence it?

src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
```diff
@@ -211,10 +213,4 @@ public Database fetchDatabase(boolean cache, DatabaseFetcherListener listener) {
 
         return fetchHostDatabase(listener);
     }
-
-    public LocalDatabaseFetcher enableCompression(boolean enableCompression)
```
1whatleytay, line 215: Thank you :DDD

src/main/java/org/hydev/mcpm/utils/ZstdException.java
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Errors during the zstd compression/decompression
+ *
+ * @author Azalea (https://github.com/hykilpikonna)
+ * @since 2022-11-22
+ */
+public class ZstdException extends IOException
+{
+    public ZstdException(Throwable cause)
```
1whatleytay, line 13: Not sure where this is used but I'm curious why we won't use `rethrow` for something like this?

src/main/java/org/hydev/mcpm/utils/ZstdUtils.java
```diff
@@ -0,0 +1,170 @@
...
+        catch (Throwable e)
+        {
+            // Any exception here would mean that it isn't supported
+            return null;
+        }
+    }
+
+    static
+    {
+        zstd = getZstdClass();
```
1whatleytay: Wish this was truly lazy but I don't really care that much.

src/main/java/org/hydev/mcpm/utils/ZstdUtils.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /**
+     * Check if zstd is supported on the current platform
+     *
+     * @return true if supported
+     */
+    public static boolean nativeSupport()
+    {
+        return zstd != null;
```
1whatleytay, line 62: `nativeSupport` implies that if this returns false there can be non-native support (this is literally a thought that passed through my smol brain). If this is false, consider renaming the method (`supported()`)? If this is true (cuz I haven't really seen anything under this) damn nice work!

src/test/java/org/hydev/mcpm/ZstdTest.java
```diff
@@ -0,0 +1,26 @@
...
+     *
+     * @param args Args (not used)
+     */
+    public static void main(String[] args) throws ZstdException
+    {
+        System.out.printf("Detected ZSTD support: %s\n", ZstdUtils.nativeSupport());
+
+        var cb = ZstdUtils.compress("Zstd compression/decompression works!".getBytes());
+        System.out.println(new String(ZstdUtils.decompress(cb)));
+    }
```
1whatleytay: Wouldn't mind moving this main method into `ZstdUtils` since that's `ZstdTest`'s sole dependency (plus `ZstdTest` is not actually a unit test like the other `*Test` classes).

src/test/java/org/hydev/mcpm/utils/ZstdUtilsTest.java
```diff
@@ -0,0 +1,60 @@
...
+ * @author Azalea (https://github.com/hykilpikonna)
+ * @since 2022-11-22
+ */
+class ZstdUtilsTest
+{
+    @Test
+    void test() throws ZstdException
+    {
+        assumeTrue(ZstdUtils.nativeSupport());
+        assertEquals(new String(decompress(compress("Hello world".getBytes()))), "Hello world");
```
1whatleytay: Worth potentially specifying encoding here for safety (`"UTF-8"`). It's a test so I guess if something goes wrong with encoding we'll know though 🙃

1whatleytay (approved):

src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
hykilpikonna: hmm it's actually not a warning. maybe the compiler optimizes it 🤔

src/main/java/org/hydev/mcpm/utils/ZstdException.java
hykilpikonna: Because we're using two different Zstd implementations here, and I want one exception to catch both of them

src/test/java/org/hydev/mcpm/ZstdTest.java
hykilpikonna: Oh ok

src/test/java/org/hydev/mcpm/utils/ZstdUtilsTest.java
hykilpikonna: Done

src/main/java/org/hydev/mcpm/client/database/fetcher/LocalDatabaseFetcher.java
1whatleytay: Looks like it! My bad. Whether or not the compiler optimizes it, I've been warned for this before (I promise it was `static final bool`), so I'm glad they fixed whatever madness was causing it.

src/main/java/org/hydev/mcpm/utils/ZstdUtils.java
hykilpikonna: > `nativeSupport` implies that if this returns false there can be non-native support (this is literally a thought that passed through my smol brain). If this is false, consider renaming the method (`supported()`)? If this is true (cuz I haven't really seen anything under this) damn nice work!

it's actually true, we default to a pure-java implementation if we don't have native support, which is slower but workable

src/main/java/org/hydev/mcpm/utils/ZstdUtils.java
hykilpikonna: Just 21 ms so it's fine

image

crawl.sh
```diff
@@ -0,0 +1,20 @@
...
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+pushd "$SCRIPT_DIR"
+
+PY="/home/azalea/.conda/envs/310/bin/python3"
+
+# Crawl packages
+$PY -m tools.run org.hydev.mcpm.server.crawlers.SpigetCrawler
+
+# Create database
+$PY -m tools.run org.hydev.mcpm.server.crawlers.spiget.CreateDatabase
```
1whatleytay, line 14: You are mixing and matching PR purposes a bit :P

1whatleytay (approved):

MstrPikachu (approved):
LGTM

PR #70, opened 2022-11-22
Use Case: Mirror select command

hykilpikonna:
Please merge #69 first.

image

PR #72, opened 2022-11-23
Use Case: Info command

hykilpikonna:
* Info command: Prints plugin information of some installed plugins to give you a better idea of what the plugin looks like.
* Includes the InfoController and InfoParser classes that handle directing plugin information and parsing info command user input respectively.
* Please merge #70 before this one

image

### DIFF
--- a/src/main/java/org/hydev/mcpm/client/commands/entries/InfoController.java
+++ b/src/main/java/org/hydev/mcpm/client/commands/entries/InfoController.java
@@ -1,0 +1,83 @@
+package org.hydev.mcpm.client.commands.entries;
+
+import com.google.common.collect.Streams;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.text.WordUtils;
+import org.hydev.mcpm.client.database.PluginTracker;
+
+import javax.annotation.Nullable;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Consumer;
+
+/**
+ * Controller for the info use case
+ *
+ * @author Azalea (https://github.com/hykilpikonna)
+ * @since 2022-11-22
+ */
+public record InfoController(PluginTracker tracker)
+{
+    private static final int LEN_LEFT = 12;
+    private static final int LEN_RIGHT = 60;
+
+    /**
+     * Format a kv pair if value isn't null, return empty string if value is null
+     *
+     * @param key Key
+     * @param rawValue Value
+     * @return Formatted pair
+     */
+    private static String formatPair(String key, @Nullable Object rawValue)
+    {
+        if (rawValue == null) return "";
+
+        String value;
+        if (rawValue instanceof Map<?,?> v) rawValue = v.keySet().stream().toList();
+        if (rawValue instanceof List<?> v) value = String.join(", ", v.stream().map(Object::toString).toList());
+        else value = rawValue.toString();
+
+        // Wrap text
+        value = String.join("\n", Arrays.stream(WordUtils.wrap(value, LEN_RIGHT).split("\n"))
+            .map(it -> StringUtils.repeat(" ", LEN_LEFT + 5) + "&6" + it).toList()).strip();
+
+        return "&7> &f" + StringUtils.rightPad(key, LEN_LEFT) + " : &6" + value + "&r\n";
+    }
+
+    /**
+     * Run the info command
+     *
+     * @param name Name of the plugin
+     * @param log Logger
+     */
+    public void info(String name, Consumer<String> log)
+    {
+        try
+        {
+            var pl = tracker.listInstalled().stream().filter(it -> it.name().equalsIgnoreCase(name)).findFirst()
+                .orElseThrow(() -> new AssertionError(String.format("Cannot find plugin '%s'", name)));
+
+            var msg = "&bPlugin Info:\n";
+            msg += formatPair("Name", "&b" + pl.name());
+            msg += formatPair("Main", pl.main());
+            msg += formatPair("Version", pl.version());
+            msg += formatPair("Description", pl.description());
+            msg += formatPair("Author", pl.author());
+            msg += formatPair("Authors", pl.authors());
+            msg += formatPair("Website", pl.website());
+            msg += formatPair("API Version", pl.apiVersion());
+            msg += formatPair("Depend", pl.depend());
+            msg += formatPair("Soft Depend", pl.softdepend());
+            msg += formatPair("Load Before", pl.loadbefore());
+            msg += formatPair("Libraries", pl.libraries());
+            msg += formatPair("Commands", pl.commands());
+            log.accept(msg);
+        }
+        catch (AssertionError e)
+        {
+            log.accept("&c" + e.getMessage());
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


src/main/java/org/hydev/mcpm/client/commands/entries/InfoController.java
```diff
@@ -0,0 +1,83 @@
...
+     * @param rawValue Value
+     * @return Formatted pair
+     */
+    private static String formatPair(String key, @Nullable Object rawValue)
+    {
+        if (rawValue == null) return "";
+
+        String value;
+        if (rawValue instanceof Map<?,?> v) rawValue = v.keySet().stream().toList();
+        if (rawValue instanceof List<?> v) value = String.join(", ", v.stream().map(Object::toString).toList());
```
1whatleytay: Inline if body is one thing, but inline if body without a newline...

```suggestion
        if (rawValue instanceof List<?> v) {
            value = String.join(", ", v.stream().map(Object::toString).toList());
        }
```

src/main/java/org/hydev/mcpm/client/commands/entries/InfoController.java
```diff
@@ -0,0 +1,83 @@
...
+            msg += formatPair("Description", pl.description());
+            msg += formatPair("Author", pl.author());
+            msg += formatPair("Authors", pl.authors());
+            msg += formatPair("Website", pl.website());
+            msg += formatPair("API Version", pl.apiVersion());
+            msg += formatPair("Depend", pl.depend());
+            msg += formatPair("Soft Depend", pl.softdepend());
+            msg += formatPair("Load Before", pl.loadbefore());
+            msg += formatPair("Libraries", pl.libraries());
+            msg += formatPair("Commands", pl.commands());
```
1whatleytay: I would count this as enough concats to justify a StringBuilder. You can also consider Collectors.joining if that's your style.

src/main/java/org/hydev/mcpm/client/commands/entries/InfoController.java
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * @param name Name of the plugin
+     * @param log Logger
+     */
+    public void info(String name, Consumer<String> log)
+    {
+        try
+        {
+            var pl = tracker.listInstalled().stream().filter(it -> it.name().equalsIgnoreCase(name)).findFirst()
+                .orElseThrow(() -> new AssertionError(String.format("Cannot find plugin '%s'", name)));
```
1whatleytay, line 78: Maybe throw together a monad? This seems like an abuse of AssertionError.

src/main/java/org/hydev/mcpm/client/commands/entries/InfoController.java
```diff
@@ -0,0 +1,83 @@
...
+    private static final int LEN_RIGHT = 60;
+
+    /**
+     * Format a kv pair if value isn't null, return empty string if value is null
+     *
+     * @param key Key
+     * @param rawValue Value
+     * @return Formatted pair
+     */
+    private static String formatPair(String key, @Nullable Object rawValue)
```
1whatleytay, line 37: Factor out Non-Null?

src/main/java/org/hydev/mcpm/client/commands/entries/InfoController.java
1whatleytay: This seems like a job for overloading.

src/main/java/org/hydev/mcpm/client/commands/entries/InfoController.java
```diff
@@ -0,0 +1,83 @@
...
+        if (rawValue == null) return "";
+
+        String value;
+        if (rawValue instanceof Map<?,?> v) rawValue = v.keySet().stream().toList();
+        if (rawValue instanceof List<?> v) value = String.join(", ", v.stream().map(Object::toString).toList());
+        else value = rawValue.toString();
+
+        // Wrap text
+        value = String.join("\n", Arrays.stream(WordUtils.wrap(value, LEN_RIGHT).split("\n"))
+            .map(it -> StringUtils.repeat(" ", LEN_LEFT + 5) + "&6" + it).toList()).strip();
```
1whatleytay: You did really good with magic numbers for `LEN_LEFT` and everything, but what does 5 represent here? I suggest making it another constant.

src/main/java/org/hydev/mcpm/client/commands/entries/InfoController.java
hykilpikonna: sorry 😭

src/main/java/org/hydev/mcpm/client/commands/entries/InfoController.java
hykilpikonna: wdym? I'm taking in nullable raw value because then I don't need to check if `plugin.something()` exists one by one

1whatleytay: This isn't necessarily bad but it does mean this method is responsible for unwrapping null. When a null value can completely switch the behaviour of a method like this, I like to move the non-null handling logic out into another method, and have another method that accepts null switch between the behaviours. I feel like this makes it harder to "opt-into" default null handling that you might not want/expect.

This method is private so it's your call, but I would still recommend moving the null-handling behaviour into its own method.

src/main/java/org/hydev/mcpm/client/commands/entries/InfoController.java
hykilpikonna: hmmm... when I finally switched everything to use StringBuilder, IntelliJ tells me I should use strings instead 🥲 

image

image

1whatleytay: Yeah, this also works (this is actually shorthand for StringBuilder if you look at bytecode). The issue is when you evaluate to a string many times via `+=` :P

hykilpikonna: hashtag # compiler is smarter than you 😂

src/main/java/org/hydev/mcpm/client/commands/entries/InfoController.java
hykilpikonna: Done

src/main/java/org/hydev/mcpm/client/commands/entries/InfoController.java
hykilpikonna: It would make it really verbose otherwise so I'll keep it like this...

I mean we can make something like `emptyIfNull(String)` that returns an empty string if the string is null, but then we would have to check if the string is empty since we don't want to display empty keys, which isn't much different from checking if the string is null. Or else we have to make something like `runIfNotNullElseReturn(String, Provider<T> run, Provider<T> default)` that takes in lambdas which is probably not worth the time

1whatleytay: I'm not going to get mad if you don't do it but...

> It would make it really verbose otherwise so I'll keep it like this...

No it wouldn't?

```java
private static String formatPairMaybe(String key, @Nullable Object rawValue) { 
  if (rawValue == null) {
    return formatPair(key, rawValue);
  } else {
    return "";
  }
}
```

I'm not asking you to do any programming gymnastics here...

hykilpikonna: Oh ok right

src/main/java/org/hydev/mcpm/client/commands/entries/InfoController.java
```diff
@@ -0,0 +1,101 @@
...
+     * @param key Key
+     * @param rawValue Value
+     * @return Formatted pair
+     */
+    private static String formatPair(String key, @Nullable Object rawValue)
+    {
+        if (rawValue == null) return "";
+
+        // If a map is passed in, only show its keys
+        if (rawValue instanceof Map<?,?> v)
```
1whatleytay, line 42: I still think it's fairly strange that you're not using overloading for this but sure... 🤔

1whatleytay (approved):

PR #75, opened 2022-11-23
[F] Fix progress bar issues with patched stdout

### LINKED ISSUE #74: Bug: Progress bar width not showing correctly in Minecraft
Since Minecraft overrides stdout with its own logger, the progress bar doesn't print correctly at the width. Maybe try printing to the original stdout with `PrintStream ps = new PrintStream(new FileOutputStream(FileDescriptor.out))`

image

hykilpikonna:
Fixes #74

### DIFF
--- a/src/main/java/org/hydev/mcpm/utils/ConsoleUtils.java
+++ b/src/main/java/org/hydev/mcpm/utils/ConsoleUtils.java
@@ -1,5 +1,7 @@
 package org.hydev.mcpm.utils;
 
+import java.io.FileDescriptor;
+import java.io.FileOutputStream;
 import java.io.PrintStream;
 
 /**
@@ -10,6 +12,7 @@ import java.io.PrintStream;
  */
 public class ConsoleUtils
 {
+    public static final PrintStream RAW_OUT = new PrintStream(new FileOutputStream(FileDescriptor.out));
     private final PrintStream out;
 
     public ConsoleUtils(PrintStream out)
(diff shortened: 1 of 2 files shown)


src/main/java/org/hydev/mcpm/utils/ConsoleUtils.java
```diff
@@ -10,6 +12,7 @@
  */
 public class ConsoleUtils
 {
+    public static final PrintStream RAW_OUT = new PrintStream(new FileOutputStream(FileDescriptor.out));
```
MstrPikachu, line 15: why this when you could say System.out

hykilpikonna: because `System.out` is patched in Minecraft. it redirects to Minecraft's logger which adds a timestamp instead of the actual System.out that just prints raw text

hykilpikonna: Or actually we can use system.err and avoid this problem

hykilpikonna: But there might be a chance that System.err is patched as well so... I guess this is the safest way

MstrPikachu: I thought System.out used FileDescriptor.out by default but ig making it explicit doesn't hurt

MstrPikachu (approved):

PR #76, opened 2022-11-23
Experimental 'augmented' local plugin trackers

kchprog:
This PR enhances the LocalPluginTracker class to make use of four fields instead of the ordinary two:

- name
- manually installed
- version ID
- plugin ID

The new class that implements this behaviour is called SuperLocalPluginTracker and allows for JSON serialization. The hope is to provide an augmented version that allows for further extension as well. This also allows for various use cases to directly get Plugin and ID information without having to look up similar plugins.


src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
```diff
@@ -0,0 +1,576 @@
...
+    Boolean isManual;
+    String versionId;
+    String pluginId;
+
+    public PluginTrackerModel(String name, Boolean isManual, String versionId, String pluginId) {
+        this.name = name;
+        this.isManual = isManual;
+        this.versionId = versionId;
+        this.pluginId = pluginId;
+    }
```
hykilpikonna: Please don't put two classes in one file, and also you can use records instead of classes to reduce the boilerplate like

```java
public record PluginTrackerModel(
    String name,
    boolean isManual,
    ...
) { }
```

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
```diff
@@ -0,0 +1,576 @@
...
+            while ((line = reader.readNext()) != null) {
+                accumulator.add(new PluginTrackerModel(line[0], Boolean.parseBoolean(line[1]), line[2], line[3]));
+            }
+            reader.close();
+            return accumulator;
+        } catch (FileNotFoundException e) {
+            return accumulator;
+        } catch (CsvException | IOException e) {
+            throw new RuntimeException(e);
+        }
```
hykilpikonna: Can you try using JSON instead of CSV? We're trying to remove the opencsv dependency because it's kind of large and one of its dependencies has known vulnerabilities

```java
// import static org.hydev.mcpm.Constants.JACKSON;
JACKSON.readValue(new File(mainLockFile), new TypeReference<>() {})
```

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
```diff
@@ -0,0 +1,576 @@
...
+        String csvFile = mainLockFile;
+        try (CSVWriter writer = new CSVWriter(new FileWriter(csvFile))) {
+            for (PluginTrackerModel pluginTrackerModel : listToSave) {
+                String[] data = { pluginTrackerModel.name, pluginTrackerModel.isManual.toString(),
+                        pluginTrackerModel.versionId, pluginTrackerModel.pluginId };
+                writer.writeNext(data);
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
```
hykilpikonna: Save json:

```java
// import static org.hydev.mcpm.Constants.JACKSON;
JACKSON.writeValue(new File(mainLockFile), listToSave);
```

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
```diff
@@ -0,0 +1,576 @@
...
+ * This class keeps track of locally installed packages
+ *
+ * @author Kevin (https://github.com/kchprog)
+ * @since 2022-09-27
+ */
+public class SuperLocalPluginTracker implements SuperPluginTracker {
+    // CSV file storing the list of manually installed plugins.
+    // Now, each row in the csv file represents something as follows:
+    // ""name", "boolean", "versionId", "pluginId"
+    private String mainLockFile = "plugins/mcpm.lock.csv";
```
hykilpikonna: If they are different implementations, please don't write to the same file

hykilpikonna:
This branch is behind... you can rebase from main to keep it up to date

image

Do this:

```sh
git checkout main
git pull
git checkout feat-commands
git pull
git rebase main
git push -f
```

hykilpikonna:
Also, SuperPluginTracker and TestLocalPluginTracker seems to be empty, did you have something that didn't go through? Currently it can't compile because `SuperLocalPluginTracker implements SuperPluginTracker` but `SuperPluginTracker` doesn't exist.

image

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
```diff
@@ -0,0 +1,619 @@
...
+    }
+}
+
+/**
+ * This class keeps track of locally installed packages
+ *
+ * @author Kevin (https://github.com/kchprog)
+ * @since 2022-09-27
+ */
+public class SuperLocalPluginTracker implements SuperPluginTracker {
```
1whatleytay, line 29: Very impressive that you were even able to do this. I expect the multiple class thing to be a compile error...

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
```diff
@@ -0,0 +1,619 @@
...
+
+        for (PluginYml pluginYml : installedPlugins) {
+            PluginTrackerModel PluginRepresentation = new PluginTrackerModel(pluginYml.name(), false,
+                    pluginYml.version(), "unknown");
+
+            // if the plugin exists in currentList and installedPlugins,
+            // add it to toAdd with its currently stored parameters and status
+
+            if (currentListName.containsKey(PluginRepresentation.name)) {
+                if (currentListName.get(PluginRepresentation.name).versionId.equals(PluginRepresentation.versionId)) {
```
1whatleytay: Nit: Use `currentListName.getOrDefault(PluginRepresentation.name, null)` and check for null instead of `containsKey` + `get`. When you use both containsKey and get you are searching the map twice.

1whatleytay: Nit: When you have a capital first letter `PluginRepresentation` it made me think you made a PluginRepresentation class lolll

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
```diff
@@ -0,0 +1,619 @@
...
+                        try {
+                            return readMeta(child).version();
+                        } catch (Exception e) {
+                            System.out.println("Error reading plugin.yml version from " + child);
+                        }
+                    }
+                }
+            } else {
+                System.out.println("Plugin with id " + name + " not found");
+                return "";
```
1whatleytay, line 412: Nit: This return does nothing.
```suggestion
```

1whatleytay (commented):
Let's get this merged

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
MstrPikachu: this seems like the real way to use the empty access modifier

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
```diff
@@ -0,0 +1,619 @@
...
+    // Now, each row in the csv file represents something as follows:
+    // ""name", "boolean", "versionId", "pluginId"
+    private String mainLockFile = "plugins/mcpm.lock.csv";
+
+    // Directory containing the plugins
+    private String pluginDirectory = "plugins";
+
+    final ObjectMapper mapper = new ObjectMapper();
+
+    /*
```
MstrPikachu:
```suggestion
    /**
```
This should be javadoc lmao

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
```diff
@@ -0,0 +1,619 @@
...
+    /**
+     * Read metadata from a plugin's jar
+     *
+     * @param jar Local plugin jar path
+     * @return Metadata
+     */
+    private PluginYml readMeta(File jar) {
+        try (PluginJarFile InstancePluginJarFile = new PluginJarFile(jar)) {
+            return InstancePluginJarFile.readPluginYaml();
+        } catch (Exception e) {
```
MstrPikachu: Please don't catch Exception, just catch the most specific exception required in case we get some obscure exception we need to debug.
```suggestion
        } catch (IOException e) {
```

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
```diff
@@ -0,0 +1,619 @@
...
+     *
+     * @param list ArrayList of PluginTrackerModel instances to save
+     *
+     */
+    private void saveJson(ArrayList<PluginTrackerModel> list) {
+        try {
+            this.mapper.writeValue(Paths.get(mainLockFile).toFile(), list);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        } catch (Exception ex) {
```
MstrPikachu, line 134: I believe the only exception here is IOException, so we also shouldn't catch Exception

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
MstrPikachu: surely naming conventions is in our checkstyle?

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
```diff
@@ -0,0 +1,619 @@
...
+            return new ArrayList<>();
+
+        // Filter only java files, return all metadata that's not null
+        return Arrays.stream(list).filter(f -> f.isFile() && f.getName().endsWith(".jar"))
+                .map(this::readMeta).filter(Objects::nonNull).toList();
+    }
+
+    /**
+     * Mark a plugin as manually installed (as opposed to a dependency)
+     * Precondition: mainLockFile is a sorted .csv file with the following format:
```
MstrPikachu: sorted is a precondition but I don't see the precondition being used or enforced anywhere. Maybe you can just remove it

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
```diff
@@ -0,0 +1,619 @@
...
+        // installedPluginsYML list
+        List<String> installedPlugins = new ArrayList<>();
+        for (PluginYml plugin : installedPluginsYml) {
+            installedPlugins.add(plugin.name());
+        }
+
+        // Get the difference between the set of manually installed plugins,
+        // the set of required dependencies, and the set of all installed plugins
+        orphanPlugins = installedPlugins;
+        orphanPlugins.removeAll(requiredDependencies);
```
MstrPikachu, line 379: if you're going to be removing, you might as well make requiredDependencies and manuallyinstalledPlugins sets

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
```diff
@@ -0,0 +1,619 @@
...
+
+        // For each plugin in the list of installed plugins, check if the version in the
+        // plugin.yml file is outdated
+        // If it is, add the plugin name to the list of outdated plugins
+        for (PluginYml plugin : installedPlugins) {
+            try {
+                if (compareVersion(plugin.name(), searchPackagesBoundary)) {
+                    outdatedPlugins.add(plugin);
+                }
+            } catch (Exception e) {
```
MstrPikachu: is it possible to catch something more specific than Exception

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
```diff
@@ -0,0 +1,619 @@
...
+                        }
+                    }
+
+                }
+            } else {
+                System.out.println("Error getting hash from server");
+                return false;
+            }
+
+        } catch (Exception e) {
```
MstrPikachu, line 507: Same thing, surely we can catch something more specific than Exception

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
```diff
@@ -0,0 +1,619 @@
...
+                    if (child.getName().equals(name)) {
+                        return child;
+                    }
+                }
+                throw new IllegalArgumentException("Plugin not found, verify whether installed.");
+            } else {
+                throw new Exception("Empty Directory.");
+            }
+
+        } catch (Exception e) {
```
MstrPikachu:
```suggestion
        } catch (IOException e) {
```

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
1whatleytay: It is and checkstyle is failing.

1whatleytay (approved):

src/main/java/org/hydev/mcpm/client/models/PluginTrackerModel.java
```diff
@@ -0,0 +1,60 @@
...
+     * Constructor for PluginTrackerModel.
+     * Instantiates a PluginTrackerModel object from a string.
+     */
+    public PluginTrackerModel(String stringRepresentation) {
+        String[] split = stringRepresentation.split(",");
+        this.name = split[0];
+        this.isManual = Boolean.parseBoolean(split[1]);
+        this.versionId = split[2];
+        this.pluginId = split[3];
+    }
```
hykilpikonna, line 39: You can remove this now since we're not parsing csv anymore

src/main/java/org/hydev/mcpm/client/models/PluginTrackerModel.java
```diff
@@ -0,0 +1,60 @@
...
+     * @param isManual  Whether the plugin is manually installed.
+     * @param versionId The version ID of the plugin.
+     * @param pluginId  The plugin ID of the plugin.
+     */
+    public PluginTrackerModel(String name, Boolean isManual, String versionId, String pluginId) {
+        this.name = name;
+        this.isManual = isManual;
+        this.versionId = versionId;
+        this.pluginId = pluginId;
+    }
```
hykilpikonna, line 27: It's better to use a record than a class here so we don't need to create getters and setters

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
```diff
@@ -0,0 +1,582 @@
...
+    public Boolean compareVersionNew(File local, PluginModel remote) {
+        try {
+            PluginYml localPlugin = readMeta(local);
+            String localVersion = localPlugin.version();
+            String remoteVersion = remote.versions().get(0).meta().version();
+            return localVersion.equals(remoteVersion);
+        } catch (Exception e) {
+            System.out.println("Error comparing versions");
+            return false;
+        }
```
hykilpikonna: You shouldn't catch all exceptions, and probably shouldn't print anywhere in the class because reporting errors to the user should be handled by the controller, and logging is actually handled differently in different environments. (e.g. if you use stderr, a Minecraft user executing the command wouldn't see the error)

src/main/java/org/hydev/mcpm/client/database/SuperLocalPluginTracker.java
```diff
@@ -0,0 +1,582 @@
...
+     * String[] data = { pluginTrackerModel.getName(),
+     * pluginTrackerModel.isManual().toString(),
+     * pluginTrackerModel.getVersionId(), pluginTrackerModel.getPluginId() };
+     * writer.writeNext(data);
+     * }
+     * } catch (IOException e) {
+     * throw new RuntimeException(e);
+     * }
+     * }
+     */
```
hykilpikonna, line 159: Please don't comment out code. Delete them. You can see the previously deleted code through git history if you ever want to reference something.

aanushkasharma (approved):

PR #78, opened 2022-11-24
[U] Apply selected mirror to LocalDatabaseFetcher and Downloader

hykilpikonna:
* Use `Supplier<URI>` instead of `URI` so that when the selected mirror updates, the URI used in LocalDatabaseFetcher and Downloader will be updated as well.


src/main/java/org/hydev/mcpm/client/arguments/CommandsFactory.java
```diff
@@ -35,18 +35,20 @@ private CommandsFactory() { }
...
-        var host = URI.create("https://mcpm.hydev.org");
-        var fetcher = new LocalDatabaseFetcher(host);
+        var fetcherListener = new ProgressBarFetcherListener();
+        var mirror = new MirrorSelector();
+        var fetcher = new LocalDatabaseFetcher(mirror.selectedMirrorSupplier());
         var tracker = new LocalPluginTracker();
-        var searcher = new SearchInteractor(fetcher);
+        var searcher = new SearchInteractor(fetcher, fetcherListener);
         var exportPluginsController = new ExportPluginsController(new ExportInteractor(tracker));
         var listController = new ListController(new ListAllInteractor(tracker));
```
1whatleytay, line 63:
```suggestion
        var listController = new ListController(new ListAllInteractor(tracker));
        
```

src/main/java/org/hydev/mcpm/client/arguments/parsers/RefreshParser.java
```diff
@@ -0,0 +1,49 @@
...
+import java.io.IOException;
+import java.util.function.Consumer;
+
+/**
+ * Parser for the refresh command
+ *
+ * @author Azalea (https://github.com/hykilpikonna)
+ * @since 2022-11-24
+ */
+public record RefreshParser(RefreshController controller) implements CommandParser
```
1whatleytay, line 16: Wouldn't mind using some more specific names at this point. `Refresh`... what? I have to run the help command to find out? :(

src/main/java/org/hydev/mcpm/client/commands/entries/RefreshController.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.io.IOException;
+
+/**
+ * Controller for the refresh command
+ *
+ * @author Azalea (https://github.com/hykilpikonna)
+ * @since 2022-11-24
+ */
+public record RefreshController(DatabaseFetcher fetcher, DatabaseFetcherListener listener, MirrorSelectBoundary mirror)
```
1whatleytay: People don't like spacing things out? T_T

```suggestion
public record RefreshController(
    DatabaseFetcher fetcher,
    DatabaseFetcherListener listener,
    MirrorSelectBoundary mirror
)
```

src/main/java/org/hydev/mcpm/client/database/mirrors/MirrorSelectBoundary.java
```diff
@@ -63,4 +65,26 @@ default List<Pair<Mirror, Integer>> pingMirrors() throws IOException
      * @param mirror A selected mirror
      */
     void setSelectedMirror(Mirror mirror) throws IOException;
+
+    /**
+     * Get the selected mirror as a URI supplier
+     *
+     * @return URI supplier function
+     */
+    default Supplier<URI> selectedMirrorSupplier()
```
1whatleytay, line 74: Don't return a `Callable` here. You're putting too much responsibility on our callee. It's trivial to wrap this method in a lambda at the call site.

1whatleytay, line 74: Also, I think this is a bit of an abuse of default methods. IMO interfaces should be behaviour-less, small enough to get a good sense of what you need to implement. You're also forgoing static dispatch.

I would recommend moving some of these out into a static class.

src/main/java/org/hydev/mcpm/utils/GeneralUtils.java
```diff
@@ -41,6 +42,33 @@ public static URI makeUrl(String url, Object... parameters)
...
+     * @param base Base url
+     * @param path Path
+     * @return Concatenated URLs
+     */
+    public static URI concatUri(String base, String path)
+    {
+        // Remove redundant slashes
+        base = StringUtils.stripEnd(base, "/");
+        path = StringUtils.stripStart(path, "/");
+        return URI.create(base + "/" + path);
```
1whatleytay, line 57: I'm crying that you didn't find a better solution than I did. I looked everywhere... I was floored that no-one provided like a good "url resolve me" method that didn't have some weird caveat T_T

1whatleytay (approved):

src/main/java/org/hydev/mcpm/client/commands/entries/RefreshController.java
```diff
@@ -0,0 +1,35 @@
...
+ * @since 2022-11-24
+ */
+public record RefreshController(
+    DatabaseFetcher fetcher,
+    DatabaseFetcherListener listener,
+    MirrorSelectBoundary mirror
+)
+{
+    /**
+     * Refresh
```
jerryzhu509: A bit more detail pls

thudoan1706 (commented):
LGTM!

jerryzhu509 (approved):

PR #79, opened 2022-11-25
[+] Pagination

hykilpikonna:
* Paginated table formatting

https://user-images.githubusercontent.com/22280294/203884161-18820b0c-dadc-4b1a-b941-0a599bc8ecbb.mp4


src/main/java/org/hydev/mcpm/client/commands/entries/PageController.java
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Controller for the page command
+ *
+ * @author Azalea (https://github.com/hykilpikonna)
+ * @since 2022-11-24
+ */
+public class PageController implements PageBoundary
+{
+    private static final File PERSIST_PATH = new File(Constants.CFG_PATH, "page-info.json");
```
1whatleytay, line 20: I believe a gateway interface is required here for adherence to clean architecture.

1whatleytay (commented):
Sounds like if someone else uses a table-related command their results will be broadcast out to everyone else (e.g. people who are browsing plugins for themselves, people who are trying to see the results of an installation, etc.) This seems like a sensible workflow to support.

We can push it out to another PR if you'd like, it seems like a fair amount of work to design a system that can keep track of the user ids of people who invoke commands. For the time being, it would help if we could allow users to opt out of pagination in-command.

src/main/java/org/hydev/mcpm/client/commands/entries/PageController.java
hykilpikonna: Maybe put the `StringStorage` interface in `feat-export-import` to use?

hykilpikonna:
> Sounds like if someone else uses a table-related command their results will be broadcast out to everyone else (e.g. people who are browsing plugins for themselves, people who are trying to see the results of an installation, etc.) This seems like a sensible workflow to support.
> 
> We can push it out to another PR if you'd like, it seems like a fair amount of work to design a system that can keep track of the user ids of people who invoke commands. For the time being, it would help if we could allow users to opt out of pagination in-command.

Yea right I should probably keep track of the user somewhere. Currently nothing in our system tracks the current user, so it would take a lot of effort to make this user-dependent.

1whatleytay (approved):

PR #82, opened 2022-11-26
Uninstall redone

aanushkasharma:
Updated uninstall usecase


src/main/java/org/hydev/mcpm/client/Downloader.java
```diff
@@ -45,19 +45,22 @@
      */
     public void downloadFile(String url, File to)
     {
+        // TODO: Implement this
```
MstrPikachu: You should delete these comments

MstrPikachu: Is this your commit even 🤔
It looks like download but your use case is uninstall

src/main/java/org/hydev/mcpm/client/interaction/ProgressBar.java
```diff
@@ -113,7 +113,7 @@ private void forceUpdate()
         if (istty) cu.curUp(activeBars.size());
         activeBars.forEach(bar -> {
             cu.eraseLine();
-            out.println(bar.toString(theme, cols));
+            out.print(bar.toString(theme, cols));
```
MstrPikachu: Why did this get changed

hykilpikonna (approved):

src/main/java/org/hydev/mcpm/client/Downloader.java
aanushkasharma: I'm not too sure why these files were even added, I was playing around with download like over a week ago just to fix the bugs and I'm guessing I accidentally committed them.

src/main/java/org/hydev/mcpm/client/interaction/ProgressBar.java
aanushkasharma: I was trying to see why downloading a single file was creating a new progress bar each time the download progressed, but if that bug has already been sorted out, you can just ignore this change

src/main/java/org/hydev/mcpm/client/injector/LocalJarFinder.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.Optional;
+
+/**
+ * Implementation that finds locally installed jars by name
+ *
+ * @param dir Plugins directory (default to "./plugins")
+ * @author Azalea (https://github.com/hykilpikonna)
+ * @since 2022-11-27
+ */
+public record LocalJarFinder(File dir) implements LocalJarBoundary
```
1whatleytay, line 18: Did this make it into here on purpose? Seems like the history for this PR is weird. Not sure if you pulled these classes in because you needed them or something.

src/main/java/org/hydev/mcpm/client/uninstall/Uninstaller.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Override
+    public UninstallResult uninstall(UninstallInput input) {
+        // 1. Unload plugin
+        // This will throw PluginNotFoundException if the plugin isn't loaded. If it is loaded,
+        // this will return the jar of the currently loaded plugin, which is the most precise.
+        // Since people may still uninstall a plugin when it's not loaded, we ignore the not
+        // found error here.
+        File jar = null;
+        if (unloader != null) {
```
1whatleytay, line 44: Might be worth marking `unloader` as `@Nullable`. I had no idea I could pass in null from the Javadoc or anything so far.

1whatleytay (approved):

PR #85, opened 2022-11-27
Update installer 

thudoan1706:


hykilpikonna (approved):

src/main/java/org/hydev/mcpm/client/DatabaseManager.java
```diff
@@ -1,35 +1,37 @@
...
 import org.hydev.mcpm.client.installer.InstallResult;
 import org.hydev.mcpm.client.installer.input.InstallInput;
 import org.hydev.mcpm.client.models.PluginVersion;
 import org.hydev.mcpm.client.models.PluginYml;
 
+import java.net.URI;
 import java.util.List;
 
 /**
  * Database API
```
MstrPikachu, line 19: if you could mention a bit more about what the Database API does that would be great

src/main/java/org/hydev/mcpm/client/DatabaseManager.java
```diff
@@ -39,16 +41,35 @@ public SearchPackagesResult getSearchResult(InstallInput input) {
...
+
+
+    /**
+     * check if plugin with given name already installed locally
      *
      * @param pluginVersion The version of the installed plugin
      * */
-    public boolean checkPluginInstalled(PluginVersion pluginVersion) {
+    public boolean checkPluginInstalledByVersion(PluginVersion pluginVersion) {
         var name = pluginVersion.meta().name();
```
MstrPikachu, line 67: name isn't used anywhere, did you forget to check the names are equal as well?

src/main/java/org/hydev/mcpm/client/arguments/CommandsFactory.java
```diff
@@ -44,9 +46,16 @@ public static List<CommandParser> baseParsers() {
         var searchController = new SearchPackagesController(searcher);
         var mirrorController = new MirrorController(new MirrorSelector());
         var infoController = new InfoController(tracker);
+        PluginLoader pluginLoader = null;
+        if (isMinecraft) {
+            pluginLoader = new PluginLoader();
+        }
+        DatabaseManager databaseManager = new DatabaseManager(tracker, searcher);
+        System.out.println(isMinecraft);
```
MstrPikachu, line 54: do we need to print here

src/main/java/org/hydev/mcpm/client/arguments/CommandsFactory.java
```diff
@@ -44,9 +46,16 @@ public static List<CommandParser> baseParsers() {
...
         var installController = new InstallController(new InstallInteractor(
             new SpigotPluginDownloader(new Downloader(), host.toString()),
-            new DatabaseManager(tracker, searcher)));
+            databaseManager, pluginLoader));
```
MstrPikachu, line 57: Can SpigotPluginDownloader handle a null pluginLoader or do we only care if it works in minecraft

src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
```diff
@@ -84,7 +96,19 @@ public InstallResult installPlugin(InstallInput installInput)
                 installPlugin(dependencyInput);
```
MstrPikachu, line 96: The recursive step doesn't use any results from the recursion. For example, if a dependency fails to be installed, the overall result would still return success if the initial plugin is a success. I can approve if we make this an issue and fix it later

src/main/java/org/hydev/mcpm/client/arguments/CommandsFactory.java
hykilpikonna: it should handle a null plugin loader 🤔

hykilpikonna: Or maybe we can create a plugin loader implementation that does nothing?

MstrPikachu (approved):
I'll make an issue

PR #87, opened 2022-11-28
avoid uninstalled dependency + add install presenter

### LINKED ISSUE #86: Not using recursive results while installing dependencies

thudoan1706:


src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
```diff
@@ -46,20 +57,25 @@ public InstallInteractor(PluginDownloader spigotPluginDownloader, DatabaseManage
...
         // 1. Search the name and get a list of plugins
         SearchPackagesResult searchResult = databaseManager.getSearchResult(installInput);
 
-
         if (searchResult.state() != SearchPackagesResult.State.SUCCESS) {
-            return new InstallResult(
+            InstallResult installResult = new InstallResult(
                     searchResult.state() == SearchPackagesResult.State.FAILED_TO_FETCH_DATABASE ?
                             InstallResult.Type.SEARCH_FAILED_TO_FETCH_DATABASE :
                             InstallResult.Type.SEARCH_INVALID_INPUT);
```
1whatleytay: Unused? This could probably just be a `State`. Still trying to figure out what you're replacing InstallResult with though.

src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
```diff
@@ -46,20 +57,25 @@ public InstallInteractor(PluginDownloader spigotPluginDownloader, DatabaseManage
...
+            resultPresenter.displayResult(installInput.name() + ": " + installResult.type().reason());
+            return false;
         }
+
         if (searchResult.plugins().isEmpty()) {
-            return new InstallResult(Type.NOT_FOUND);
+            InstallResult installResult = new InstallResult(Type.NOT_FOUND);
```
1whatleytay: ^^ If you're going to get rid of InstallResult, maybe you should delete it and see if that cleans some stuff up.

src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
```diff
@@ -46,20 +57,25 @@ public InstallInteractor(PluginDownloader spigotPluginDownloader, DatabaseManage
...
         SearchPackagesResult searchResult = databaseManager.getSearchResult(installInput);
 
-
         if (searchResult.state() != SearchPackagesResult.State.SUCCESS) {
-            return new InstallResult(
+            InstallResult installResult = new InstallResult(
                     searchResult.state() == SearchPackagesResult.State.FAILED_TO_FETCH_DATABASE ?
                             InstallResult.Type.SEARCH_FAILED_TO_FETCH_DATABASE :
                             InstallResult.Type.SEARCH_INVALID_INPUT);
+            resultPresenter.displayResult(installInput.name() + ": " + installResult.type().reason());
```
1whatleytay: BTW I would pass the presenter the raw stuff (ex. just the name() and the type(), let it handle the formatting!)

src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
```diff
@@ -118,6 +145,13 @@ public InstallResult installPlugin(InstallInput installInput)
      */
     public static void main(String[] args) {
         new File(FILEPATH).mkdirs();
+        Consumer<String> log = new Consumer<>() {
+            @Override
+            public void accept(String s) {
+                System.out.println(s);
+            }
+        };
```
1whatleytay, line 169: IntelliJ might offer a simpler solution too!

```suggestion
        Consumer<String> log = s -> System.out.println(s);
```

src/main/java/org/hydev/mcpm/client/installer/presenter/InstallPresenter.java
```diff
@@ -0,0 +1,27 @@
...
+     * @param log log string to the console
+     */
+
+    public InstallPresenter(Consumer<String> log) {
+        this.log = log;
+    }
+
+    @Override
+    public void displayResult(String message) {
+        this.log.accept(message);
```
1whatleytay: This is pretty bare bones. I would recommend returning an InstallResult object with everything you need to print. Let the presenter turn that into a nice string!

src/main/java/org/hydev/mcpm/client/installer/presenter/ResultPresenter.java
```diff
@@ -0,0 +1,14 @@
+package org.hydev.mcpm.client.installer.presenter;
+
+/**
+ * Interface for presenting the state of plugins
+ *
+ */
+public interface ResultPresenter {
```
1whatleytay: This doesn't present all results... right?

```suggestion
public interface InstallResultPresenter {
```

1whatleytay (approved):

1whatleytay (approved):

src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
```diff
@@ -46,20 +57,25 @@ public InstallInteractor(PluginDownloader spigotPluginDownloader, DatabaseManage
      * @param installInput: the plugin submitted by the user for installing
      * */
     @Override
-    public InstallResult installPlugin(InstallInput installInput)
+    public boolean installPlugin(InstallInput installInput,
+                                       ResultPresenter resultPresenter)
```
MstrPikachu: In clean architecture, your presenter should depend on the use case, not the other way around

src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
```diff
@@ -75,40 +91,51 @@ public InstallResult installPlugin(InstallInput installInput)
...
-                installPlugin(dependencyInput);
-            }
-        }
-
-        if (pluginLoader != null && installInput.load()) {
-            try {
-                pluginLoader.loadPlugin(installInput.name());
-            } catch (PluginNotFoundException e) {
-                return new InstallResult(Type.SUCCESS_INSTALLED_AND_FAIL_LOADED);
+                installPlugin(dependencyInput, resultPresenter);
```
MstrPikachu: Depending on what you're returning, you probably want to use this result somehow. The first example I gave in the PR was just a proof of concept, because you still want to return an InstallResult to pass to the presenter later.
But the logic is still that you want to combine your InstallResults for the dependencies into some kind of final result

src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
1whatleytay: Wait why is this me I don't remember changing this. Isn't this from the installer PR?

thudoan1706: oops, thank you! I will fix it by passing the presenter to the interactor constructor instead of use case method.

src/main/java/org/hydev/mcpm/client/installer/presenter/InstallPresenter.java
thudoan1706: Gotcha!

src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
MstrPikachu: what I mean is that like before, you should return an InstallResult so the result can go the the controller, who passes it to the presenter

src/main/java/org/hydev/mcpm/client/installer/presenter/ResultPresenter.java
thudoan1706: I thought of having install, uninstall and update share the same interface with results since they all seem to have State. It also works if all of them pass the presenter to the use case, but I will leave it InstallResultPresenter for now XD

src/main/java/org/hydev/mcpm/client/installer/InstallInteractor.java
thudoan1706: Ohh, I am actually thinking of changing the InstallResult returned type to boolean, letting the presenter handle the job showing the state of manual install input and its dependency. If I return the InstallInput to the controller, I won't be able to know what happened to the dependency, whether or not they have been installed. I think this way works well for testing too ><

hykilpikonna (approved):

PR #89, opened 2022-11-28
Warnings cleanup and Project Reorganization

1whatleytay:
I think it's time to rip off the bandaid! This PR's end goal is to reorganize the project and get rid of as many warnings as possible. Some things this PR does:

 - We remove author comments. TA feedback specifically mentioned these, they also cause 1 warning per raw GitHub link we mention, so we'll be using VC instead for the time being.
 - Reorganize packages, all classes of a related certain layer (Entity/Application/Interface/Frameworks) should be in their own package. Right now package names and interfaces are a bit all over the place, so this PRs goal is to fix this.
 
 There's a new project structure! I'm down for feedback, but this is the way that made more sense to me. **GET YOUR PRs MERGED ASAP** so I'm the one who has to deal with conflicts. There could be still some classes I forgot to work with, LMK if I missed anything:

For all `client` classes:

 - `arguments`: Classes related to ArgsParser. Contains ArgsParser and some parsing utilities.
   - `parsers`: **All** `CommandParser` implementing classes. These classes should talk to controllers in `client.commands.controllers`.
 - `commands`: A side-effect less package for controllers/presenters. No printing to stdout here.
   - `controllers`: **All** use-case controllers. They should interface with presenters in `client.commands.presenters`.
   - `presenters`: Presenter boundaries. Since controllers should not really write to stdout, we want the actual implementation to be in `client.display.presenters`.
 - `database`: Interface classes related to interacting with a database. Actual classes that make net requests or interact with the file system should be in `client.local`.
   - `fetcher`: Database fetcher related classes.
   - `mirrors`: Mirror model and interface classes.
   - `tracker`: PluginTracker model and interface classes.
 - `display`: This package is renamed from `interaction` (since we have a ton of "interactors" in our project that aren't in this package). This is the home for presenters and other classes that write to stdout.
   - `presenters`: Put the implementation for presenters here!
   - `progress`: Put ProgressBar related classes here!
 - `local`: **ALL INTERACTION WITH NET/FS** should be done here! Let's keep this contained. I called it `local` since most net related classes seem to also interact with the "local" filesystem. Not totally sure how that works.
 - `models`: I believe entities should live here. So Plugin/Database related classes.
 - `export`: Export use-case related classes here. Anything that doesn't write to stdout or deals with interface classes is OK.
 - `injector`: Hot reload related classes here. This directory is a bit privileged to talk with Spigot (which is a side effect). Up to you if you want to split that out into another directory.
 - `installer`: Installer related classes. Again, no writing to stdout.
 - `list`: List-locally installed plugin related classes here (Kevin's use-case).
 - `matcher`: Not a real use case but its used for the updating system (e.g. related versions). Unsure if it's still needed but I'm keeping it around in it's own directory for now.
- `search`: Search use-case related classes here!
- `uninstall`: Uninstall related use-case classes here! Again no stdout or file-system if possible please. Might want to make like a PluginDestination interface and work with installer on this one.
- `updater`: Update related use-case classes here!


projectavi:
Wow this is so cool! Imagine if it were written in a non esoteric programming language like Kara!
Highly Recommend: https://github.com/1whatleytay/kara!

1whatleytay:
Thank you kind soul.

CONTRIBUTE.md
```diff
@@ -72,15 +72,15 @@ The downloadable Spigot plugins and their meta info are stored on our server, ho
...
+### How to setup a mirror
 
 The MCPRS server is hosted with a plain file server that supports both http and rsync. The official server is hosted using Nginx, but any file server with such compatibility would work. You can follow one of the approaches below to set up a mirror.
 
 After setting up a mirror, if you want to add it to our mirror list, you can submit a pull request to this repo editing the [mirrorlist.yml](mirrorlist.yml) file.
 
 #### Setup Mirror using Docker Compose
 
-For convenience, we created a docker image so that you can set up a mirror using Docker. It will automatically set up:
+For convenience, we created a docker image so that you can setup a mirror using Docker. It will automatically set up:
```
MstrPikachu: https://www.merriam-webster.com/dictionary/setup
setup is a noun, set up is the verb

1whatleytay: This was actually marked by the IntelliJ spellcheck :P Do you want me to change it back?

src/main/java/org/hydev/mcpm/client/updater/UpdateInteractor.java
```diff
@@ -20,7 +17,11 @@
 import java.util.Map;
 import java.util.stream.Collectors;
 
-import static org.hydev.mcpm.client.updater.UpdateOutcome.State.*;
+import static org.hydev.mcpm.client.updater.UpdateOutcome.State.MISMATCHED;
+import static org.hydev.mcpm.client.updater.UpdateOutcome.State.NETWORK_ERROR;
+import static org.hydev.mcpm.client.updater.UpdateOutcome.State.NOT_INSTALLED;
+import static org.hydev.mcpm.client.updater.UpdateOutcome.State.UPDATED;
+import static org.hydev.mcpm.client.updater.UpdateOutcome.State.UP_TO_DATE;
```
MstrPikachu, line 25: is there a warning for importing *

MstrPikachu (commented):
looks fine I think

src/main/java/org/hydev/mcpm/client/updater/UpdateInteractor.java
1whatleytay: I don't know. IntelliJ just cleaned this up on its own when Azalea hit the cleanup code button. Typically, check style doesn't like star imports, but it hasn't been complaining lately.

project_plan_dev.md
```diff
@@ -1,104 +0,0 @@
-# Project Planning and Development with Github
```
hykilpikonna, line 1: you can also remove images/

src/main/java/org/hydev/mcpm/client/arguments/parsers/CommandConfigurator.java
```diff
@@ -0,0 +1,39 @@
+package org.hydev.mcpm.client.arguments.parsers;
+
+import net.sourceforge.argparse4j.inf.Subparser;
+
+/**
+ * Provides an interface for providing Command details and setting up a Subparser object.
+ */
+public interface CommandConfigurator {
```
hykilpikonna, line 8: nice

hykilpikonna (approved):

CONTRIBUTE.md
MstrPikachu: IntelliJ has skill issue I guess we can change it later

MstrPikachu (approved):
1 spelling mistake is ok :P

PR #91, opened 2022-12-02
Add update tests

1whatleytay:
This PR is a test patch for the update use case.
 - This test brings the coverage of the update package from 33% to 96%.
 - It also adds some utility mock classes for testing use cases that involve LocalPluginTracker.
 - It _does not_ remove LocalPluginTracker, that will be done in another PR.
 - It does fix some update related bugs that were still present.

Some new classes:
 - `MockInstaller`: Mocks the InstallBoundary interface, saves the successful "installations" into a list that can be checked.
 - `MockPluginTracker`: Holds PluginTracker data in a list that can be checked and edited. It does not persist, and can be modified and checked after a test completes.
 - `UpdateInteractorTest`: Contains some tests for the update use case! Yay.

### DIFF
--- a/src/main/java/org/hydev/mcpm/client/DatabaseManager.java
+++ b/src/main/java/org/hydev/mcpm/client/DatabaseManager.java
@@ -44,6 +44,10 @@ public class DatabaseManager {
      * @param pluginName The version of the installed plugin
      * */
     public boolean checkPluginInstalledByName(String pluginName) {
+        if (!localPluginTracker.findIfInLockByName(pluginName)) {
+            return false;
+        }
+
         List<PluginYml> pluginInstalled = localPluginTracker.listInstalled();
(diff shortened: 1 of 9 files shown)


hykilpikonna (approved):

src/main/java/org/hydev/mcpm/client/local/SuperLocalPluginTracker.java
```diff
@@ -22,7 +22,7 @@
  * This class keeps track of locally installed packages
  */
 public class SuperLocalPluginTracker implements SuperPluginTracker {
-    // CSV file storing the list of manually installed plugins.
+    // CSV file storing the list of manualbly installed plugins.
```
MstrPikachu:
```suggestion
    // CSV file storing the list of manually installed plugins.
```

MstrPikachu (approved):
surely is fine

PR #94, opened 2022-12-05
Search Presenter

jerryzhu509:
In the previous iteration of SearchController, we implicitly had it act as a presenter by having it write to Consumer<String> log. To comply with clean architecture (e.g. single responsibility principle), we separate the presenting functionality out into a SearchResultPresenter interface and its concrete implementation SearchPresenter, as per the clean architecture diagram. It also warranted changes to how we initiate the search use case in ArgsParser, which I adjusted.

Removed author tags and added to javadoc, per TA comments.


MstrPikachu (approved):
looks good

1whatleytay (approved):
Good stuff!

PR #95, opened 2022-12-07
Add controller and parser tests

1whatleytay:
This PR introduces tests for almost all use case controllers and parsers. It also reorganizes CommandsFactory (as part of the TA feedback). These tests can be found in the `org.hydev.mcpm.client.arguments` package.

Some key classes:
 - `MockUseCaseClasses` implement the boundary for our use cases, so we can test just the parser and the controller without interfering with the actual use cases.
 - `UseCaseTest` classes in the arguments package have a list of some basic tests that test for basic behaviour (correct parsing, objects are passing correctly). I tried to avoid checking for presentation logic because I understand that's up in the air.
 - `InteractorFactory` and `InteractoryFactory` boundary. Since these factories require state, I decided to go with the abstract factory pattern to uhh, properly decide how the boundaries should be filled.
 - I also went with an abstract factory pattern with Controllers in the `ControllerFactory` and `ControllerFactoryBoundary`, even if it makes much less sense there. Here I do it because each Controller might have a different preferred way of being instantiated (e.g. do we really need to instantiate it with ControllerFactoryBoundary). It also offers convenience with fields to store Factory data.
 - `ParserFactory` which creates the parser list (which can then be passed to ArgsParserFactory).

Hopefully this can give us some nice coverage. We are now at 36% lines covered from 21% project wide.
Screenshot 2022-12-07 at 1 42 40 AM


src/main/java/org/hydev/mcpm/client/arguments/ControllerFactory.java
```diff
@@ -0,0 +1,93 @@
...
+    @Override
+    public ReloadController reloadController() {
+        return new ReloadController(boundary.reloadBoundary());
+    }
+
+    @Override
+    public UnloadController unloadController() {
+        return new UnloadController(boundary.unloadBoundary());
+    }
+}
```
jerryzhu509, line 93: That's amazing! W for code smells

jerryzhu509 (approved):
This PR is truly marvelous. I can't believe you single-handedly made factories for ArgsParsers, Controllers, Interactors,.. and made tests for EVERY parser. Hats off to you 

Looks good as far as I can tell - and the tests pass so it's probably fine.

hykilpikonna (approved):

PR #99, opened 2022-12-07
Search Without Static Storage

jerryzhu509:
Per TA comments, I added comments to SearchInteractorTest that better illustrate what each function tests.
Before, I implicitly assumed that we will be working with just the same database, hence making the search maps static. However, this assumption is quite limiting, and ran into problems when tests are ran. Now, I basically just made them non-static and instead, reinstantiate the searchers when needed.

I also mixed up the order of assertEquals for expected and actual. Fixed now.


1whatleytay (approved):

MstrPikachu (approved):
Everything is good
The only thing is that this assumes that the database does not change between instances of SearchInteractor
In this case we can skip all the initialization stuff in SearchInteractor#search and move it into the constructor

PR #100, opened 2022-12-07
removing files in uninstall

aanushkasharma:
This PR adds two key classes to help the Uninstall use case adhere to Clean Architecture and SOLID:
 - The FileRemove interface, which allows the Uninstall use case to make requests to delete plugins without interfacing with the filesystem directly.
 - And the RemovePlugin class that implements the FileRemove interface and performs the correct operations to delete files.


src/main/java/org/hydev/mcpm/client/uninstall/RemoveFile.java
```diff
@@ -0,0 +1,36 @@
...
+public class RemoveFile implements FileRemove{
+    private final LocalJarBoundary jarFinder;
+
+    public RemoveFile(LocalJarBoundary jarFinder) {
+        this.jarFinder = jarFinder;
+    }
+
+    public int removeFile(String pluginName) {
+        // 2. If it isn't loaded, find the plugin jar file in local file system
+        // (This will throw PluginNotFoundException when a plugin of the name in the file system
```
kchprog: This comment is inaccurate. Should read "If a plugin is not found, return 1"

kchprog (approved):

